PR #13, opened 2026-09-03
test(gateway): cover the unpair fix as shipped, not as re-written

BANG88:
`#5` fixed "unpairing a gateway that is gone looks ignored", and its test passed without ever running the code that was fixed. This adds coverage that does, and reports what the fix actually does on a device.

## What the old test could not see

`gateway-unpair.test.ts` defined its own `reviveOrForgive` and asserted against that. The copy had drifted from `removeRecord` in three ways:

- it kept a **local `REVOKE_BUDGET_MS`**, so changing the store's number broke nothing;
- it decided forgiveness with a **local regex** rather than `describeGatewayFailure`;
- it **cleared the race's timer**, which the store does not.

And the on-screen half had no coverage at all — nothing knew the button says "Unpairing…", and nothing knew the Cancel beside it stops taking taps.

The middle one matters most. The store invents `new Error('Timed out waiting for the server.')` and then asks `describeGatewayFailure` what it is; that function classifies **by matching the words "timed out" in the message text**. Rewriting that sentence turns an unreachable gateway back into an error toast and a record that will not go away — the reported bug, restored silently. Confirmed by mutation: renaming the sentence to "The server took too long." now fails the suite, where before it failed nothing.

## What is here

- **`gateway-unpair.test.ts` drives `removeRecord` itself**, with only SecureStore and the network stubbed. Six cases: a gateway that answers is asked then forgotten; one that never answers is given up on and forgotten anyway, inside the budget; a network failure is forgiven the same way; a 401 and a 500 each still block the removal; an unknown record asks nobody anything.
- **`src/lib/unpair-action.ts`** holds the control's two-button state, and `UnpairAction` now reads it instead of spreading `pending` across four props. It is a view model rather than a rendering test because `bun test` cannot parse React Native's Flow entry point — and because the failure worth guarding is not a missing spinner, it is **a spinner with a live Cancel beside it**, which looks finished and is not.
- **`bun:test`'s ambient shim** gains `mock.module` and `test`'s timeout argument.

## Device evidence

Run against a scratch gateway on its own port, with the app on a simulator/emulator served by Metro from this branch. All three states, both platforms:

| | iOS | Android |
|---|---|---|
| Armed (both buttons live) | ✅ | ✅ |
| Reachable gateway: spinner, record removed, **token actually revoked** | ✅ | ✅ |
| Unresponsive gateway: spinner, both buttons inert, record removed | ✅ | ✅ |

The unresponsive case was produced by `SIGSTOP`ing the gateway — the socket still accepts, nothing ever answers, which is what a machine that has gone away looks like. Sampling the screen every ~0.2 s on iOS: armed until t≈0.55 s, "Unpairing…" with a live spinner from t≈0.65 s, record gone by t≈5.3 s. **The wait is ~4.5 s** (the 4 s budget plus the local write behind it), against the ~16 s it replaced.

Two things worth knowing that the commit message does not claim:

- **Giving up is silent.** The row simply disappears; there is no message saying the server could not be reached. That matches the code (the timeout is *forgiven*, so `unpair` takes the success path) and it is a real improvement over a 16 s stall, but a reader is never told the gateway still holds the token.
- **The gateway does still hold it.** After a timed-out unpair the device's token remains server-side — the documented trade-off. It is visible and revocable from another paired device, which the Android screenshots show.

## Gates

`oxfmt --check`, `oxlint --deny-warnings`, `tsc --noEmit`, `bun test src` — all clean (2524 pass, 0 fail).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/settings-servers.tsx
+++ b/src/components/settings-servers.tsx
@@ -24,6 +24,7 @@ import { feedback } from '@/lib/feedback';
 import { loadPairedDevices, revokePairedDevice, type PairedDevice } from '@/lib/gateway-client';
 import type { GatewayRecord } from '@/lib/gateway-storage';
 import { fadeIn, fadeOut, listLayout, timing } from '@/lib/motion';
+import { acceptsCancel, acceptsConfirm, unpairView } from '@/lib/unpair-action';
 import { useRenderTally } from '@/lib/render-tally';
 import { describeGatewayFailure } from '@/lib/network-error';
 import { normalizeGatewayUrl } from '@/lib/pairing';
@@ -592,6 +593,7 @@ function UnpairAction({ label, onUnpair }: { label: string; onUnpair: () => Prom
   const theme = useThemeTokens();
   const [armed, setArmed] = useState(false);
   const [pending, setPending] = useState(false);
+  const view = unpairView({ armed, pending });
 
   // Collapsing the row and coming back must not leave a primed destructive
   // button waiting.
@@ -602,7 +604,7 @@ function UnpairAction({ label, onUnpair }: { label: string; onUnpair: () => Prom
   // tap looks ignored -- the row sits there, unchanged, until the record
   // vanishes on its own. Say what is happening instead.
   async function confirm() {
-    if (pending) return;
+    if (!acceptsConfirm({ armed, pending })) return;
     setPending(true);
     try {
       await onUnpair();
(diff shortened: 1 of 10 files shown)


BANG88:
Rebased onto `d4626e1` and fixed the `mock.module` leak you hit. Force-pushed.

**The defect was real and worse than a red suite.** `mock.module('@/lib/gateway-storage', …)` is process-wide, so my seven-function stub deleted `saveGateway`, `normalizeGatewaySshTunnel` and the rest from every other suite for the rest of the run — `gateway-storage.test.ts` (which landed after this branch started) went red in a full run while passing alone.

Took your route: **stopped replacing our own module.** The real `gateway-storage` now runs underneath the store test over a fake keychain and node's crypto, exactly as `gateway-storage.test.ts` already did. That is strictly stronger than what it replaces — "the record is gone" is now asserted against storage that really sealed it and really removed it, and the only remaining stub is the far side of the network, which is the input the test exists to control.

**It also exposed an older ordering hazard, which is why the first attempt still failed.** Four suites each registered their own *partial* `expo-secure-store`, and whichever is in place the first time a module imports it is the one that module keeps for the whole run. So a fake missing a method fails whichever **other** suite reaches that code path — `removeGateway` deletes its keys when the last record goes, and none of the partial fakes had `deleteItemAsync`. All four carry the full surface now, and the two gateway suites share one vault (`src/lib/__tests__/gateway-vault.ts`) so neither can end up resetting an object nothing is bound to.

**Verified in one process, not per file:**

- `bun test src` — **2589 pass, 2 skip, 0 fail** (2591 across 98 files)
- Same result in either directory order (`src/lib src/stores` and `src/stores src/lib`)
- Neither suite passes vacuously, confirmed by mutation **in a full run**: breaking `normalizeGatewaySshTunnel` fails the storage suite (2 fail); renaming the store's `Timed out waiting for the server.` fails the unpair suite (1 fail)
- `oxfmt --check`, `oxlint --deny-warnings`, `tsc --noEmit` all clean

One note: `tsc` was initially red on `server-terminal-workspace.tsx` for the `/sessions` route from #12 — stale generated `.expo/types/router.d.ts` in my worktree, not a code issue. Regenerating the route types cleared it; nothing in the repo needed changing.

The device evidence in the PR description is unchanged — that verification ran against the same shipped `settings-servers.tsx` and `gateway-connection.ts`, which this commit does not touch.

PR #41, opened 2026-09-05
feat(quick-actions): four tiles where five rows and their sentences were

BANG88:
The quick actions sheet opened on five full-width rows, each with a title and a
sentence under it, and that block alone was most of a phone screen before the
saved shortcuts had started. The sentences were describing verbs their own
titles had already named. This replaces them with one row of four tiles and
tightens the two lists underneath.

## The new layout

**A sticky top.** Handle, header and the tile row are one node at
`stickyHeaderIndices={[0]}`, so the sheet's own verbs do not scroll away from a
list that runs to thirty-odd rows against a real gateway. This is not a new
idiom: the panels sheet has pinned its handle-and-header node the same way since
card #633, with the platform-only grabber inside the node because a sticky index
has to be constant. The node reclaims the content container's paddings with
negative margins, the same trick and for the same reason -- otherwise rows slide
through the gap beside it on their way up.

**Four tiles.** New terminal, New group, Simulator, Browser -- an icon over a
word, `flex: 1` each in a row with `LADDER.gap` between them, which is how the
on-screen keyboard states a key width and what a Pad's 560-point measure needs
anyway. Each tile is a `GlassChrome face="sheet"` carrying the shape
(`appChrome.radius.control`, `borderCurve: 'continuous'`, `overflow: 'hidden'`)
with a `PressableScale` filling it, which is what every other control on an
opaque sheet in this app wears. 82 points tall: 12 above and below a 20-point
glyph, a 4-point lead, and two lines' worth of label height whether the label
uses one line or two -- without that last part a tile whose label wraps centres
a taller stack and lifts its glyph above its neighbours' by half a line.

Three tile states, each drawn rather than implied:

* **busy** -- the spinner stands exactly where the icon was, so the row does not
  change height under the thumb.
* **disabled** -- `appChrome.opacity.disabled`, ink at `textSubtle`, the reason
  in the tile's own `accessibilityLabel` *and* in one muted caption under the
  row. A tile is a word wide, so a per-tile sentence does not fit; a dimmed
  control with no reason anywhere is the state this sheet has always refused to
  draw (see `webServiceBlockedByTunnel` in `quick-actions.ts`). The caption is
  inside the sticky node, so it stays with the tiles it explains.
* **selected** -- `primary` ink on the simulator tile while a Pad is actually
  showing a simulator beside the terminal. The only tile whose press leaves
  something on screen to report back, and the third and last place this sheet
  spends the accent.

A connection that offers neither machine-scoped tile (a plain-HTTP LAN gateway,
or the offline demo) draws the two it has at half width apiece rather than
leaving two holes in a four-column grid. That is the common case, not an edge
one, so it is worth saying: the row is even in both shapes.

**Not tiles.** New task keeps a full-width row under them -- it is the only one
of the sheet's verbs that hands over to a form with three questions on it rather
than doing a thing, and a label that has to say so is a sentence. Stop keeps its
own tinted surface, and moved from above the tiles to below them: a surface that
comes and goes must not sit above the things that stay, now that the things that
stay do not move.

**Edit** moved from the SAVED PROMPTS heading to the sheet header, beside Close,
as a pencil that becomes a check. That is where this app puts a sheet's own
actions -- panels, Files and simfarm all keep theirs there -- and it is honest
about the mode's scope: Edit reveals a second line under every saved row, a
delete beside each one, and a form at the foot of the sheet, so a button sitting
inside one section's heading was claiming the mode belonged to that section.

**The two lists.** Every row on the sheet is now 48 points tall whether it
carries one line or two (4 points of padding above and below a 20/17 stack),
where it was 56 for a single second line and 73 for a wrapped one -- so a list of
thirty agent commands had two rhythms in it depending on how long each
description happened to be. 48 still clears the 44-point minimum, and a mistap
here sends something to a live pane, so it does not go lower.

* *Saved prompts* are title-only. A saved prompt is prose the reader wrote and
  then named, so the value line was the title again at greater length
  ("Summarize progress" over "Summarize the current progress and remaining
  work."). A saved *command* is `git branch --show-current` under "List
  branches", which the title cannot say, so terminal mode keeps its monospace
  second line -- the typeface is still the taxonomy. Edit brings the prompt back,
  which is the one moment its exact wording is what is being decided about.
* *Agent commands* keep the shape they had -- a monospaced command, a muted
  argument hint beside it, a description under it -- but the description is
  capped at one line instead of two. Not grouped: the gateway returns them in
  its own order, the source annotation already says where each came from, and
  splitting thirty rows under two headings adds two headings without shortening
  the list.

## Space reclaimed

Measured from the stylesheets on a 390x844 phone, agent mode, the five-row card
the sheet used to open with (no New task, no Stop) and the offline demo's five
saved prompts and six agent commands:

| | before | after |
|---|---|---|
| header block | 46.2 | 42.2 |
| the sheet's verbs | 315.3 (5 rows) | 82 (4 tiles) |
| SAVED PROMPTS heading | 32 (the Edit button's 32pt tap target set it) | 20 |
| top of sheet to first saved prompt | **433.5** | **192.2** |
| the five saved prompts | 332.3 | 241.3 |
| the six agent commands | 354.7 | 289.7 |
| whole sheet | **1160.5** | **763.2** |

**241.3 points reclaimed above the saved list (56%), 397.3 over the whole sheet
(34%).** On the first screenful -- roughly 750 points of a full-detent form sheet
on an 844-point phone -- the fold moves from inside the saved prompts to inside
the agent commands: before, the header, the five action rows and four of the five
saved prompts, with AGENT COMMANDS entirely below the fold; after, the header,
the tiles, all five saved prompts and five of the six agent commands. On the
874-point device this was verified on, the whole sheet fits with room to spare.

## What "Render as text" took with it

The sheet was the only way into that reading, so removing the row alone would
have left a mode nobody could enter -- and, for an install upgraded from the old
`agentTerminalMode` switch, one nobody could leave. The two other entry points
had already gone: `PaneViewModeControl` was written but has never been mounted
anywhere in this repository's history, and Settings' "Agent panes open in" has
been behind `AGENT_DEFAULT_VIEW_SETTING_ENABLED = false` since 2026-07-27. So
the mode goes, not just the row:

* `PaneViewMode` loses `'text'`; `availablePaneViewModes` can no longer return
  it and `isPaneViewMode('text')` is now false.
* `storedAgentDefaultView` loses the `agentTerminalMode === false -> 'text'`
  migration. A persisted `agentDefaultView: 'text'` now reads as nothing stored,
  so such an install opens its agent panes on the terminal like every other one.
  `agentDefaultView` itself stays -- it is still what a pane opens in, and chat
  still uses it.
* `server-terminal-workspace.tsx` loses `agentOutput` and the three branches it
  chose: the `mergeTerminalWindow` fold in `applyPaneOutput`, the
  `TerminalBoundary` reset key's `agent`/`terminal` suffix, and the output
  format, which is now the module constant `PANE_OUTPUT_FORMAT = 'ansi'`. The
  constant keeps the name because it is also the head of `outputShape`, the key a
  remembered window is filed under -- a window filed before this change is still
  filed under `ansi:`, so nobody's scrollback is orphaned.
* `terminal-output.tsx` loses its `mode` fork and the cross-fade between the two
  surfaces. It is now one `SkiaTerminal`; the wrapper stays because the gateway
  zoom effect is its own reason to exist.
* `agent-markdown-output.tsx` is deleted -- it drew that reading and nothing
  else. `createMarkdownStyle` and its `syntaxColors` helper are read by the chat
  transcript's blocks and by the asset viewer, neither of which is a terminal,
  so they move to `src/lib/markdown-style.ts`: two pure functions over the
  theme's colours, no React.
* `lib/agent-chrome.ts` and `lib/transcript-history-scroll.ts` go with it. Each
  had exactly one consumer, and it was that component. `scripts/test-terminal.ts`
  loses the agent-chrome block for the same reason -- it was asserting on a
  module nothing rendered.
* `mergeTerminalWindow` stays exported and tested in `src/terminal/history.ts`,
  which is out of scope for this change. It now has no app call site.

Net: 36 files, +768 / -1761.

## Strings

`bun run i18n` reports **0 missing in all eight locales** (694 messages).

Two ids are new, both translated in all eight:

| id | en | de | fr | es | pt |
|---|---|---|---|---|---|
| `Browser` | Browser | Browser | Navigateur | Navegador | Navegador |
| the tunnel caption | see below | Der SSH-Tunnel tragt den Port des Gateways und keinen anderen, ... | Le tunnel SSH transporte le port de la Gateway et aucun autre : ... | El tunel SSH lleva el puerto de la Gateway y ningun otro, ... | O tunel SSH transporta a porta da Gateway e mais nenhuma, ... |

The `zh-TW`, `ja` and `ko` translations of both are in their catalogs. The
caption in full:

> The SSH tunnel carries the Gateway's port and no other, so the simulator and
> the browser cannot be reached from here. Connect over the local network or
> Tailscale.

The tile labels reuse existing translated ids wherever one existed: `New
terminal`, `New group` and `Simulator` (already carried by the simfarm sheet).
`Preview a simulator`, `Open in your browser` and `Hide the simulator` are reused
too, but as the tiles' accessibility labels -- the tile shows one word and the
screen reader hears the whole verb. That is also why the labels are short:
"Ouvrir dans le navigateur" is three lines in a quarter of a 390-point phone, and
"Navigateur" is one. Checked on device in `de`, `fr`, `pt` and `ja`; the longest
wrap any of them produces is two lines ("Nouveau"/"terminal").

Fifteen ids are dropped in all eight locales, and `lingui extract --clean` took
them out of every catalog:

* the five tiles' old descriptions -- `Add a terminal beside this one, and go to
  it.`, `Start a group of its own, and go to it.`, `Watch this machine's iOS,
  Android or WeChat simulator.`, `A dev server or preview running on this
  machine.`, plus the Pad pair `Show it beside the terminal, and watch it
  redraw.` / `Give the whole width back to the terminal.`
* the two old disabled-row sentences, replaced by the one caption above:
  `Reached the same way as a browser, so the same tunnel limit applies.` and
  `The SSH tunnel carries the Gateway's port and no other, so a browser here
  cannot reach the rest of the machine. Connect over the local network or
  Tailscale to open one.`
* the removed reading's own words -- `Render as text`, `Render as terminal`,
  `Agent output as plain text. Easier to read, no colour.`, `Agent output
  reflowed for reading. Colour is lost.`, `Switch to the text view`, `Text`
* `Edit`, whose button became a glyph. `Done` survives; other screens use it.

## Verified

On `muqun-ios` (iPhone 17 Pro simulator, 402x874) and `muqun_android`
(1080x2400 / 393dp), against the offline demo workspace, in both themes:

* each tile does what its row did -- New terminal and New group create and close
  the sheet, Simulator opens `/simfarm`, Browser opens `/web-service`, both by
  `replace` as before
* the dimmed tiles carry the tunnel reason in the caption under the row and in
  their own accessibility labels; the tile row exposes `New terminal`, `New
  group`, `Preview a simulator`, `Open in your browser` to the accessibility
  tree
* saved prompts still send and close the sheet; agent commands still send
  (`/clear`) or prefill the composer with a trailing space (`/model`, which has
  an argument hint)
* Edit shows each prompt's value on a second monospace line, a delete beside
  each row, and the NEW SHORTCUT form
* with the command list padded to thirty rows, the header and tiles stay pinned
  while the list scrolls under them, on both platforms

The Pad branch (`isPadLayout`) was exercised by forcing it on the phone, which
covers `padStickyTop`'s negative-margin reclaim against `padContent`'s wider
gutter; the 560-point centred measure was not seen on a real iPad, since booting
one was out of scope for this change's device budget.

## Gates

```
npx tsc --noEmit      clean
bun run lint          oxlint --deny-warnings, clean
bun run format:check  All matched files use the correct format
bun test src          2766 pass, 2 skip, 0 fail (2768 across 103 files)
bun run i18n          694 messages, 0 missing in de/es/fr/ja/ko/pt/zh-TW
bash scripts/e2e.sh   matches main -- see below
```

`bash scripts/e2e.sh --device emulator-5554`, run against a freshly booted
`muqun_android` with the app served from Metro, and then run again unchanged
against `c0c1bf7` from a second worktree on the same emulator:

| flow | `c0c1bf7` | this branch |
|---|---|---|
| demo-tour | pass | pass |
| pairing-manual | pass | fail, then **pass** on two retries |
| terminal-interactions | fail | fail |
| ssh | fail | fail |
| file-mentions | fail | fail |
| attachments-ui | pass | pass |
| artifacts | pass | pass |
| settings | fail | fail |
| away-digest | pass | pass |
| pad-workspace | fail | pass→fail identically |
| slash-commands | pass | pass |

Five flows fail identically on `main` and on this branch, with the same
assertions -- `Switch to zsh`, `Open Demo shell`, `Type into this editor`
(twice) and `Contact us.*`. They are pre-existing on this harness and none of
them is in a surface this change touches.

The sixth, `pairing-manual`, failed once here on `Assertion is false: "Install
command copied" is visible` and passed on `main`. It is flake, not a
regression: it also passed on this branch in an earlier run, and it passes on
two consecutive re-runs of the flow on this branch, reaching that exact
assertion both times. Nothing in this diff goes near pairing or the clipboard.

`demo-tour` is the flow that opens and closes this sheet, and it passes on both.

So the suite is **no worse than `main`** -- but it is not green on `main`
either, and this change does not fix that.

Rebased onto `c0c1bf7`; the one conflict was `terminal-output.tsx`, where that
commit's new `paneRows` / `paneCursorColumn` / `paneCursorRow` props are kept and
passed straight through to the single `SkiaTerminal` this file now renders.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


BANG88:
The e2e baseline I promised is in, and the answer is that this branch matches
`main`.

Same emulator (`muqun_android`, booted once and left alone), same harness, the
second run from a throwaway worktree at `c0c1bf7` with nothing else changed:

| flow | `c0c1bf7` | this branch |
|---|---|---|
| demo-tour | pass | pass |
| pairing-manual | pass | fail once, pass on two retries |
| terminal-interactions | fail | fail |
| ssh | fail | fail |
| file-mentions | fail | fail |
| attachments-ui | pass | pass |
| artifacts | pass | pass |
| settings | fail | fail |
| away-digest | pass | pass |
| pad-workspace | fail | fail |
| slash-commands | pass | pass |

Five of the six failures are identical on both, down to the assertion that
trips: `Switch to zsh`, `Open Demo shell`, `Type into this editor` (in both
`file-mentions` and `pad-workspace`) and `Contact us.*`. Those are pre-existing
and this change does not touch any of those surfaces.

The sixth is `pairing-manual`, which failed here on `Assertion is false:
"Install command copied" is visible` while passing on `main`. That one is flake:
it passed on this branch in an earlier run of the same suite, and re-running the
flow on this branch twice in a row passes both times, reaching that exact
assertion. The diff goes nowhere near pairing or the clipboard.

`demo-tour` is the flow that opens and closes this sheet, and it passes on both
sides.

So: no regression, and the sheet's own flow is green -- but the suite is not
green on `main` either, and this PR does not fix that. Worth a card of its own.

PR #63, opened 2026-09-13
docs: design for multi-agent dispatch from a phone (review, not implementation)

BANG88:
## Revision 3 — the direction, and a companion document

Rebased onto `main`. Still docs only; no app or gateway source is touched.

Two files now: `docs/multi-agent-dispatch.md` and the new
`docs/agent-capability-matrix.md`.

**The decision this revision records: the app does not build an orchestrator.**
Orchestration is done by an **architect agent** running in a Herdr pane, driven
by a skill in the cross-agent `.agents/skills/` convention (`muqun-architect`).
It splits a brief, routes each part to the agent whose *agent-specific*
capability fits — image generation to Codex `$imagegen` / Amp Painter / Cursor,
long-context multi-file refactors to Claude Code, sandboxed destructive
experiments to an agent with a real OS sandbox — starts one pane and worktree
per part named `<brief>/<part>`, and collects a `RESULT.md` per part into a
summary.

The app's share is four things, and none of them is a scheduler: a one-tap
**"Architect"** composer target that injects the capability catalogue as a JSON
block via `POST /tasks`; a **task-group view grouped on the naming prefix**;
per-agent observation, approval and interrupt **exactly as today**; and results
through the **existing artifacts view**. No progress percentages, no
"completed" — the `AGENTS.md` rules are restated in the section itself, because
a group view is where they erode.

### What changed in the design document

- New top section **Direction (2026-09-14)** — the decision, what the architect
  does, the `<brief>/<part>` name and the `RESULT.md` contract, what the app
  builds, what it must not do, and the phase list.
- New section **Capability catalogue** — static and maintained in the Gateway in
  the shape `agents.json` already has (per kind: agent-specific capabilities,
  approval-prompt shape, detectable-in-tmux, verified-against version), plus
  runtime "installed kinds" from `GET /api/agents/catalog`. It argues why
  runtime "what can you do" queries are not viable: three of twelve agents
  answer machine-readably, the Gateway never builds a command line to ask from,
  and asking costs a pane and a turn. It carries the agent-specific versus
  commodity split, and three corrected facts — Codex's skill is **`$imagegen`**
  (model `gpt-image-2`, plan-gated, burns limits 3–5×), **Claude Code has no
  image generation**, and prompt delivery is a **TTY paste** (`src/main.rs:6508`)
  so architect↔worker communication is text and files by construction.
- **§7 is re-tiered, not rewritten.** P0: wire the `agent_collaboration` gate on
  both sides (§6.1 — a fix PR is in progress), unify `AGENT_KINDS` vs
  `HERDR_AGENTS` (21 vs 14, `gemini` missing from the second), `agent_args` on
  `POST /spawn`, and the two approval-detector blind spots (Aider's inline y/n;
  Amp never prompts, so an Amp pane can never be `Blocked`). P1: §7.4, §7.8,
  §7.9 and the task-group view. P2: the architect skill, the target, the
  `RESULT.md` contract, the artifacts entry, and an image-generation end-to-end
  demo. P3: §7.6 and §7.7. **Deferred: §7.1–§7.3** — hand-authored brief parts
  and the sequential dispatch sheet, "unvalidated need; superseded by the
  architect agent". §8 is unchanged.
- §1–§6 keep every finding and citation. The only other change to them is
  `oxfmt`'s markdown formatting.

### The new companion document

`docs/agent-capability-matrix.md` is the research this routes on, cleaned up for
the repository: what the Gateway supports (Part 1, `file:line` against
muqun-gateway `dccdd67`), twelve per-agent tables read from vendor documentation
on 2026-09-14 (Part 2), and the implications (Part 3). Its header says plainly
that it is a dated snapshot that will drift, and how to refresh a row.

`bun run format` + `bun run format:check` pass. No source changed, so the E2E
gate does not apply to this PR.

---

## Design for review — not an implementation

One file, `docs/multi-agent-dispatch.md`. No app or gateway source is touched.

### Question 1: an agent inside Herdr is told, in one sentence, to create several panes and put a different agent in each. Is the pane creation Herdr's capability, or the skill file's?

**Herdr's, entirely. The skill file adds no capability at all.**

A process in a Herdr pane inherits `HERDR_ENV=1`, `HERDR_PANE_ID`, `HERDR_TAB_ID`,
`HERDR_WORKSPACE_ID`, `HERDR_SOCKET_PATH` and `HERDR_BIN_PATH`, and `SKILL.md`
describes the ambient situation rather than something it provides: "the `herdr`
binary in `PATH` talks to the current session." `pane split` and `agent start`
are ordinary subcommands of that binary, with no precondition beyond a running
server and a pane at its prompt. Nothing is installed to unlock them.

The skill is, in the docs' own words, "a Markdown instruction file for agents"
that "**tells** an agent to use the `herdr` CLI when `HERDR_ENV=1` is set." The
Socket API settles the layering: Agent skill / CLI wrappers / Raw socket API, and
"**The layers share the same control surface.**" Remove the skill and the agent
keeps every ability and loses only the knowledge.

What the skill genuinely changes is behaviour quality — don't steal focus, parse
ids from JSON, don't re-send an ambiguous prompt, escalate approval dialogs
rather than answering them — and it *narrows* engagement via its frontmatter. A
competence and safety upgrade, not an authorisation.

**On authorisation: none is documented anywhere.** The Socket API documents the
transport and socket paths in full and never mentions authentication, tokens, or
a permission model; the minimal request carries no credential. The effective
boundary appears to be filesystem reachability of a Unix socket under the user's
own config directory. The docs do not state this as a security model and do not
state socket file permissions — §2.3.3 says so plainly rather than inferring. If
it matters for a deployment decision it needs verifying against the
implementation. `HERDR_ENV=1` is an indicator Herdr sets, not a server-side
check; the skill's guardrail is advice to the agent ("should stop").

§2.3 carries every quotation with its page URL, and §2.3.4 is a one-table summary
of the line.

### Question 2: does our app already show panes and agents that Herdr creates?

**Partly, and better than expected — we show the panes, we just never show that anything happened.**

From the code (§4). If an agent splits three panes and starts three agents while
the reader has the terminal workspace open and foregrounded:

| Surface | Shows them? | When |
|---|---|---|
| Pane strip, same tab | Yes | ~250 ms via SSE; ≤12 s poll fallback |
| Pane strip, another tab or workspace | No | — |
| Visible terminal | Unchanged, correctly | — |
| Assignment roster | Yes, if `instance_id` is present | Same as strip |
| Panels sheet | Yes, everywhere | Only on open / pull-to-refresh |
| Home-screen card | Only after next opening that server | Hours; dimmed after 5 min |
| Layout / adjacency | Never | — |
| Any announcement | None | — |
| App backgrounded | Nothing updates | Until foreground |

The plumbing already handles panes it did not create — `use-pane-events.ts`
subscribes to `pane.created` and `pane.agent_detected`, and
`workspace-selection.ts:29-51` correctly refuses to follow Herdr's `focused` to a
new pane. Two real gaps: `tabPanes` filters to the selected tab, so work in
another tab is invisible on that screen; and the app has no layout model at all
(`layout_updated` is deliberately unsubscribed, one terminal mounted at a time),
so it shows *membership*, never *arrangement*. It cannot say three panes belong
to one brief.

### What changed in this revision

The first commit answered a mis-framed question — whether one Herdr command could
fan work out. That answer stands (§2.2: no, and every socket agent method is
single-target), but it was the wrong thing to lead with. This revision rewrites
the document around the two questions above, adds §4 from the app code, and adds
two proposals that follow: an *unattributed agent* state for agents that appear
without a `CollaborationTask` (§7.8, since a skill-equipped agent can fan out
from ordinary prompt text with no involvement from us), and making the panels
sheet live (§7.9).

### Still the most important finding about our own code

§6.1 — the capability gate our `AGENTS.md` mandates is not wired on either side.
`agent_collaboration` sits in a static array no gateway handler branches on, and
`collaborationAvailability` (which is both the capability check *and* the Herdr
0.9.0+ floor) has no call site in `src/` outside its test. The real gate is
`gatewaySupportsAgentSpawn`, which returns `false` unconditionally because
`AGENT_SPAWN_SHIPPED = false`. Argued as its own card ahead of this work.

Also worth a look: §6.3, where "Open terminal" on the phone calls `agent.focus`
and therefore clears the Done badge in the operator's Herdr TUI, because `done`
is defined as "idle but not yet marked seen" and reads do not mark seen.

Open questions in §9.


ryuhzk:
**Consolidated recommendation: task-focused mobile workflow, with Gateway-owned records and agent-owned planning**

This comment consolidates the proposed App experience, the Gateway work needed to support it, and the maintenance implications. It is a design recommendation for review, not an implementation or a claim that these capabilities already ship. The PR files remain unchanged. The related open Gateway PR is [muqun-gateway#26](https://github.com/osuki-dev/muqun-gateway/pull/26); its immediate scope is capability reporting, with the task service proposed as follow-up work.

I agree with the central direction in revision 3: the App should not implement an AI orchestrator. I would change the primary entry point, replace naming-based task relationships with explicit records, and introduce a small task service in Gateway. That service records and executes explicit operations; the lead agent still decides how to divide and coordinate the work.

**1. Start with the user workflow**

The main loop should be **describe the goal → handle decisions → review the result**. Pane management remains available as a supporting tool.

- **Create:** choose a project, write the goal, attach reference files, and use “Automatic” by default. Show the selected machine/project clearly. An advanced option selects a particular assistant. Reuse the existing composer and attachment pipeline rather than creating another input implementation.
- **Work:** stay on a task detail screen showing the lead assistant, any delegated work, decisions needing attention, and submitted artifacts. Starting a worker must not navigate to its terminal or change the current reading position.
- **Clarify:** a normal follow-up goes to the lead assistant. The user can explicitly address a worker, inspect its terminal, answer its actual approval prompt, or interrupt it. Display the destination before submission; never silently broadcast a follow-up.
- **Review:** show the preview/files/diff, a concise explanation, and validation evidence with known gaps. Offer “Request changes” and “Mark accepted.” Acceptance must not implicitly merge, publish, delete a worktree, or stop an agent.

“Automatic” should first use one assistant. That assistant delegates only when independent work or a different available tool justifies it. A small edit should not require a separate architect plus several workers. The Architect is a role within the workflow, not a mandatory concept the user must understand before sending a request.

For example, “Refresh the homepage artwork and adapt the page colors” might produce image work and page work. “Change this heading” should usually stay with one assistant. Delegation should have bounded concurrency, depth, and task count, and respect the user's allowed agents and execution settings. Do not display estimated cost without actual supporting data.

A useful task screen could read:

```text
Homepage visual update                         Project: website

Lead assistant
Image work assigned; page changes await the selected image.

Work
Artwork             Codex          Working
Page styling        Claude         Waiting for artwork
Final review        Lead           Not started

Needs your attention
Image service authorization
[View request]

Results
No result submitted yet

[Add instructions]                              [More]
```

“Waiting for artwork” must come from an explicit task dependency/report. “Working” comes from observed agent state. Neither should be inferred from the other. No progress percentages; no task-success claim from an idle agent.

**2. Keep ownership explicit**

| Owner | Responsibilities |
| --- | --- |
| App | Goal entry, task views, explicit user decisions, pinned output snapshots, local reading/history preferences |
| Lead agent and skill | Understand the goal, choose workers, describe dependencies, request dispatch, review submissions, integrate changes, summarize results |
| Gateway task service | Durable task/attempt records, identity binding, validated operations, delivery receipts, result references, recovery and client synchronization |
| Terminal backend adapter | Native terminal/process operations, input/output, topology and available lifecycle evidence |

Gateway must not choose models, invent subtasks, evaluate whether code is correct, or automatically advance a dependency graph. It can validate an explicit dependency and refuse a premature dispatch; the lead agent decides when to request the next operation. This is a bounded task service, not a general workflow engine.

The phone must be able to disconnect without becoming the scheduler. If the lead exits, the recorded assignments and results remain readable. Replacing the lead is an explicit new attempt supplied with the recorded context; it does not silently restart or resend work to existing workers.

**3. Use explicit identities, not labels**

The `<brief>/<part>` convention is useful display text, but should not determine group membership. Repeated briefs, renamed branches, moved panes and restarted agents make names insufficient.

Proposed minimum records:

| Record | Important fields and meaning |
| --- | --- |
| Task | `task_id`, optional `parent_task_id`, session/project scope, title, brief, explicit dependency IDs, revision |
| Attempt | `attempt_id`, `task_id`, selected agent kind, opaque agent instance identity, current terminal locator, workspace/worktree reference; a replacement creates another attempt |
| Delivery | `delivery_id`, `attempt_id`, request key, payload digest, operation type, receipt/outcome and timestamps; each follow-up has its own record |
| Result submission | `submission_id`, `task_id`, `attempt_id`, artifact references, content digest/version, summary and validation evidence/provenance |

Scope every relationship by the Gateway installation and session. Backend IDs remain opaque. A pane is a locator, not the identity of the assigned process/conversation. Keep prior attempts as history rather than rebinding them to a replacement occupant.

Keep these dimensions separate:

- Delivery: prepared, submitting, acknowledged, refused, or unconfirmed.
- Observed agent state: working, idle, blocked, unknown, with source and observation time.
- Result: no submission, submitted for review, changes requested, accepted by a user.

A submitted result is a claim backed by linked evidence, not automatic proof of correctness. User acceptance is distinct from the device-local “mark reviewed” action on assignment history. Offline history review/removal must still work without contacting or stopping an agent.

**4. Build Gateway around the existing boundary**

Gateway already has a useful hexagonal boundary: [`TerminalBackend`](https://github.com/osuki-dev/muqun-gateway/blob/b6bca766858545d9628bcc261e933134f528c624/src/backend/model.rs) and separate Herdr/tmux adapters. Preserve it.

Implement incrementally:

1. Extract reusable startup and delivery use cases from `main.rs` behind small application services. Existing HTTP handlers remain compatible. Do not expand `main.rs` into a second scheduler or create parallel implementations for each backend.
2. Add a task domain module, a repository interface, and one durable local storage implementation. Use transactions for related records and request-key uniqueness. An embedded SQLite store is a reasonable candidate to evaluate; no hosted database, Redis queue, or additional daemon is required by this design. Record schema versions and test migrations before choosing the final storage format.
3. Add task query and explicit mutation endpoints. Illustrative resources are task list/detail, attempts, deliveries, result submissions, and user acceptance. Exact route names need an API review: the existing `POST /api/sessions/{sid}/tasks` already means “prepare a workspace/worktree and start an agent,” so do not silently redefine its response or partial-success behavior.
4. Add a narrow local CLI/control adapter for the architect skill. Creating children, requesting a launch, checking a receipt and submitting a result must call the same application services as the App. Otherwise direct Herdr calls would bypass the very records the App needs.
5. Publish committed task changes over SSE with a revision/cursor and a snapshot recovery path. Reconnect must recover missed facts; a notification must not be the only record of a result.

The local agent interface should be scoped to the assigned task/session and allowed operations. Do not give a worker a paired-device credential or access to unrelated projects. A skill explains behavior; it is not an authorization boundary. A local socket is also not automatically process isolation against arbitrary same-user programs.

Agents started directly through Herdr or tmux remain visible as observed, unattributed agents. Never manufacture a task relationship from their names. Explicit adoption can be a later feature, conditional on verifiable identity.

**5. Treat delivery uncertainty as a first-class outcome**

Persist the intended operation before dispatch. Serialize writes for the same target, validate the expected instance as close to the native operation as possible, and record the acknowledgment afterward.

A client request key prevents duplicate execution of the same recorded request. The same key with a different payload must be rejected. Repeating a request or querying its receipt must return its recorded outcome rather than automatically submitting the prompt again.

This still does **not** provide exactly-once TTY delivery. Gateway can crash after writing to the terminal but before recording the acknowledgment. On recovery, that operation becomes unconfirmed; it must not be automatically replayed. Distinguish a documented refusal before any input was sent from a timeout with unknown effects.

Likewise, a fresh lookup followed by a pane write is not an atomic instance-bound send. The adapter contract must expose what it can actually guarantee. Strong delivery requires an instance-aware native operation or an owned process/input channel that cannot fall through to a replacement shell. If that guarantee is unavailable, retain the limitation explicitly and do not advertise stronger support.

Record partial resources created during startup. Failure after creating a pane/worktree must not tell the user that nothing happened, nor trigger broad rollback that deletes usable work.

**6. Support tmux and Herdr through capability levels**

| Surface | Herdr | tmux |
| --- | --- | --- |
| Ordinary terminal use | Preserve | Preserve |
| Observed external agents | Use available identity/status evidence | Show available evidence; unknown remains unknown |
| Managed task attempts | Use verified adapter identity and delivery behavior | Requires additional managed-launch identity/readiness/delivery work |
| Results and task history | Gateway-owned | Same Gateway-owned model once enabled |

Current tmux `start_agent` recognizes the foreground program and returns `instance_id: None`; that does not establish an identity-bound, prompt-ready conversation. See the [current adapter](https://github.com/osuki-dev/muqun-gateway/blob/b6bca766858545d9628bcc261e933134f528c624/src/backend/tmux.rs).

For a later managed-tmux implementation, bind a random launch generation to an owned process lifecycle and the specific tmux server/pane incarnation. A wrapper or supervisor may provide lifecycle/exit evidence. A PID, pane option, pane name, or generated UUID alone is insufficient: process exit, pane respawn, server restart and Gateway restart must invalidate or conservatively suspend the binding. Input must never fall through to the shell after the managed process exits. Prove readiness separately from process presence.

Start with observed terminal use on tmux, then enable managed task features only after those contracts pass. Do not change the existing Herdr-only `agent_collaboration` meaning merely to make tmux appear supported. Use additive, documented capabilities for new task operations, checked for the selected session and enforced server-side.

**7. Make results and capability data maintainable**

`RESULT.md` remains useful human-readable content. Add explicit result registration containing the originating attempt, validated artifact references and evidence. The App can reuse its Markdown/image viewers without parsing prose to reconstruct task state.

The current asset scanner is bounded and skips directories including `dist`, `build`, and hidden directories. Therefore “write the result and it will appear in artifacts” is not a reliable delivery contract. Reuse validated artifact serving, but register task outputs explicitly. A result reference should identify a content version; later file edits must not silently change an already reviewed submission. Report missing/deleted artifacts honestly and define retention before offering automatic cleanup.

Use isolated worktrees for concurrent writers. The lead/integration attempt owns combining changes and final validation. A worktree does not isolate ports, services or databases; account for shared resources when deciding whether work is independent. Read-only work need not allocate a checkout. Cleanup is separate from acceptance and must preserve dirty or unmerged work.

Maintain one agent registry with separate fields for launch support, executable discovery, session integration, status detection, approval handling and optional tools. The existing `AGENT_KINDS` and `HERDR_AGENTS` answer different questions; unify ownership without assuming every kind has equal support.

A binary on PATH is not proof of authentication, entitlement, a configured tool, or a capability in the selected session. Keep versioned static knowledge, local discovery and verified runtime evidence distinct. Unknown remains unknown. Do not encode permanent rules such as “vendor X always does images” or “vendor Y always handles large refactors.” The lead should receive a versioned catalogue snapshot and route against available tools and user preferences. Bound and cache inexpensive probes; do not start agents just to ask what they can do.

**8. Keep ongoing maintenance bounded**

- Store task facts and result references, not a continuously persisted copy of every terminal transcript. Set retention and pagination limits; do not silently evict active records. Separate history cleanup from terminal/worktree destruction.
- Keep one task state transition implementation shared by HTTP and the local agent interface. Vendor behavior belongs in profiles/adapters; backend transport details stay out of App components and application services.
- Use additive API changes, explicit capabilities, schema migrations and tolerant clients. Preserve legacy task/spawn envelopes and ordinary terminal behavior for older Apps and Gateways.
- Make SSE an optimization over queryable state. Reconnect from a snapshot when a cursor expires. Update attention indicators without replacing a pinned output/result snapshot; refresh remains the reader's choice.
- Keep per-device reading state local. Shared task changes use revisions to reject conflicting edits. Do not silently promote legacy device-local dispatch history into authoritative server tasks.
- Correlate diagnostics with task/attempt/delivery IDs. Log bounded error categories and timing, not prompts, credentials or full terminal output by default.
- Avoid a plugin framework, distributed queue, general DAG scheduler or broad event-sourcing rewrite in the first version. Extract modules when a concrete task use case needs them.

**9. Delivery order and verification**

1. **Compatibility foundation:** selected-session gating, truthful status, explicit instance checks, and current multi-agent visibility/history fixes.
2. **One-task vertical slice:** create one task, start one assistant, submit a follow-up, reconnect, register a result and request changes/accept. Establish durable records and receipts here.
3. **Delegation:** the lead uses the same API to create child tasks and isolated writing attempts; App renders explicit parent relationships. Add limits and lead-loss recovery.
4. **Managed tmux parity:** prove lifecycle/readiness/input guarantees before enabling equivalent task controls.
5. **Refinements:** richer routing profiles, proactive notifications, evidence-backed previews and explicit cleanup.

Each implementation needs the repository gates. Startup/delivery changes additionally need a real paired App → Gateway → agent run on an isolated session and dedicated QA device. Test both supported platforms before claiming coverage. Test backend contracts for replacement, pane reuse/moves, server/Gateway restart, lost acknowledgment, repeated request keys, partial startup, lead exit, stale approvals and artifact mutation/removal. Offline demo flows prove UI behavior, not actual delivery. No implementation tests were run for this comment-only review.

**Source freshness and proposed changes to revision 3**

The source review used App `be4293e` and Gateway `b6bca76` (the head of open Gateway PR #26, not a claim about merged main or a released build); it does not establish what is installed on any user's machine. The PR's Gateway research is pinned to the earlier `dccdd67`.

At the reviewed Gateway revision, [session capability computation](https://github.com/osuki-dev/muqun-gateway/blob/b6bca766858545d9628bcc261e933134f528c624/src/main.rs#L346) already exists. The original statement that collaboration is only in an unconditional static array is stale for this checkout. That does not prove every operation is gated correctly. In the reviewed App, [`collaborationAvailability`](https://github.com/osuki-dev/muqun-app/blob/be4293e944a5a810a132ae8dd7712aa6bf56c386/src/lib/agent-collaboration.ts#L74) still has no production call site, and [`AGENT_SPAWN_SHIPPED`](https://github.com/osuki-dev/muqun-app/blob/be4293e944a5a810a132ae8dd7712aa6bf56c386/src/lib/agent-spawn.ts#L51) is false. Dispatch history is [device-local](https://github.com/osuki-dev/muqun-app/blob/be4293e944a5a810a132ae8dd7712aa6bf56c386/src/stores/agent-collaboration.ts); Gateway's existing create-task workflow is not the durable task model proposed above.

The recommended changes are therefore: **“Automatic” task entry instead of a mandatory Architect target; explicit task/attempt relationships instead of prefix grouping; registered results instead of filename discovery; and a small durable Gateway task service shared by both terminal adapters.** Agent-owned planning, ordinary terminal access, individual approvals, honest uncertainty and reader-controlled snapshots all remain central.

ryuhzk:
**Architecture handoff index — current defects, module boundaries and execution order**

The design and implementation plan are now attached to the relevant PRs as comments. No repository files, PR titles or PR descriptions were modified.

- [Overall App workflow and module architecture](https://github.com/osuki-dev/muqun-app/pull/63#issuecomment-5659013436): goal entry, a single lead by default, explicit delegation, result review, ownership boundaries and maintainability.
- [Current App collaboration review and APP work packages](https://github.com/osuki-dev/muqun-app/pull/77#issuecomment-5659062620): verified source-level defects, regression cases, affected modules and App completion criteria.
- [Gateway design, protocol baseline and GW work packages](https://github.com/osuki-dev/muqun-gateway/pull/26#issuecomment-5659070299): persistence, API/service boundaries, operation receipts, recovery, result registration, scoped local agent control, adapter contracts and managed tmux feasibility.

**How the existing feature relates to the proposal**

`agent_collaboration` is currently a manual assignment/observation capability. It sends instructions to an existing or newly started agent and keeps device-local dispatch history. The planned module builds durable tasks, attempts, explicit child relationships, result submissions and user acceptance above those execution primitives.

They are related layers of one feature area. Reuse the delivery services, backend adapters, composer and viewers. Do not implement a second prompt transport, reinterpret idle as task success, or rename local assignment history into a server task model without adding the missing contracts.

The existing capability fixes should remain focused. They do not have to implement the entire task architecture. They do need to satisfy their own compatibility and delivery requirements.

**Review findings to resolve first**

1. App #77 gates the visible toggle but can retain an armed assignment after a Herdr → tmux session switch. The send path then uses that retained selection with the current session. The fix must cover selection scope, all entry paths and the dispatch boundary, not just rendering.
2. Current history controls disappear for reviewed/superseded/departed-agent assignments because the only UI drops the history partition. Persisted history must remain manageable offline independently of live agents.
3. Gateway's existing agent send API has no expected-instance precondition. Client preflight checks help, but do not make native delivery atomic with identity validation. Record this existing limitation and implement/prove the stronger contract before advertising it.
4. The PRs still report required device and paired delivery evidence as outstanding. Health probes and unit tests are not a substitute.

The first architecture comment cited App main `be4293e`, where the helper was unwired and the spawn constant remained false. The subsequent review checked App #77 head `78abdda`, which addresses those static gate problems; the retained-state bypass is the remaining interaction defect. Gateway `b6bca76` is open PR #26's head, not a statement about a released build. Follow those pinned scopes when interpreting the comments.

**Execution sequence once the design is confirmed**

| Stage | Gateway packages | App packages | Outcome |
| --- | --- | --- | --- |
| Repair current collaboration | GW-C1 | APP-C1, APP-C2, APP-C3 | Correct eligibility, truthful outcomes and accessible history |
| Freeze the module contract | GW-T1 | Review schemas and capability fallback | Stable task/attempt/operation/result interfaces and migration plan |
| Deliver one usable task loop | GW-T2, GW-T3 | APP-T1 | Create, receive an actual response, follow up, reconnect, submit a result, request changes/accept |
| Add delegation | GW-T4 | APP-T2 | Lead-created child work with explicit records and scoped authority |
| Extend backend support | GW-T5, then GW-T6 if feasible | Consume verified capabilities | Managed tmux task operations with demonstrated identity/readiness/input guarantees |

The package definitions in the linked comments include affected modules, acceptance evidence and dependencies. Contract-dependent packages should not proceed against independently invented schemas.

**Recommended decisions to confirm together**

Adopt task-first Automatic entry; one assistant unless delegation is useful; explicit IDs instead of name-prefix relationships; local transactional Gateway storage; additive task APIs preserving existing endpoints; separate delivery/process/result/review states; one shared service for App and local agent operations; explicit artifact registration; Herdr first and proven managed-tmux support later; no automatic retry of uncertain input or implicit publish/cleanup.

These remain recommendations until the design is confirmed. After confirmation, implementation agents should use the named package and protocol baseline directly. Material changes to ownership, delivery guarantees, persistence or compatibility should return for architectural review rather than being silently decided inside a feature PR.

Each handoff must report pinned revisions, contract changes, actual test results, limitations and missing paired/device evidence. Worktree isolation is for concurrent code changes, not a substitute for isolating shared services. Keep explicit trust decisions, user-controlled snapshots, ordinary terminal access and release authorization intact throughout implementation.

ryuhzk:
**UI interaction specification, rationale and safety requirements**

This extends the [architecture handoff](https://github.com/osuki-dev/muqun-app/pull/63#issuecomment-5659075917). It is a proposed interaction contract for confirmation and implementation, not a claim that the UI has been built or usability-tested. All changes in this update are PR comments only.

The design goal is to make three questions easy to answer at every step: **What work am I looking at? Who receives my next action? What will that action change?** The usability rationale below is an architectural hypothesis to validate through representative tasks, not evidence from user research already performed.

**UI-01 — Navigation and information architecture**

On a phone, add a Tasks entry within the existing project/server experience. Do not replace the terminal or require a new global navigation system in the first increment.

```text
Project / machine
  ├─ Tasks → Task detail → Worker detail → Terminal
  │                    ├─ Decision detail
  │                    └─ Result detail
  └─ Existing terminal workspace
```

The task list has a compact “Needs your attention” section and a stable recent-task list. Each row shows title, project/machine, last meaningful recorded event and a textual state. Tapping a row opens that task; it never answers a prompt. A visible “New task” action opens composition. If there are no tasks, show that action and one short example, not an agent-management tutorial.

Do not reorder rows underneath an active touch, keyboard focus or screen-reader focus. While reading, accumulate “Updates available” rather than moving the selected item. Do not use an agent's `done` flag to label a task complete. “Result submitted” requires an actual submission; “Accepted” requires a recorded user review.

On a tablet, use a task list plus detail pane with the same selection and back-navigation semantics. Preserve the chosen task and its scroll position across layout changes. Resizing must not initiate a task or change the active terminal.

**Why:** tasks preserve the user's intention across agent restarts and multiple workers. A terminal remains valuable for inspecting raw behavior, but making it the primary task container forces the user to remember which pane contains which responsibility.

**UI-02 — Create a task with one clear commit point**

```text
New task                                         Close

Project: website · Development Mac                 Change
Assistant: Automatic                              Change
Delegation: Within this project                   Settings

What would you like done?
[Goal editor                                      ]
[                                                 ]

[Attach]   reference.png · Ready

                                      [Start task]
```

- Opening New task does not start an agent. The project may be prefilled from the explicit current project; the machine remains visible. With no project context, require selection rather than silently choosing a machine.
- Reuse the existing editor, attachment queue and validation. Keyboard Enter inserts a newline; the labeled Start task action is the commit point. Keyboard shortcuts must have the same preconditions as the button.
- Automatic means one lead assistant selected from available, allowed profiles. Show its resolved identity before committing the launch. It does not mean “enable unattended approvals” or “always start multiple agents.”
- Delegation settings expose an allowed agent set, maximum concurrent workers and project scope. Show the effective execution/approval profile. Existing project preferences can be reused; expanding them requires an explicit change, not a hidden flag in an agent command line. Unsupported enforcement must not be presented as a guarantee.
- Changing the machine/project preserves the written goal as a draft, invalidates the resolved assistant and attachment destination, and requires re-selection/re-upload as appropriate. Do not silently forward files staged for one machine to another.
- Upload progress is real byte progress and may show a percentage; task progress must not. Failed attachments have inline Retry/Remove controls. While Start task is waiting for uploads, label it “Uploading attachments…” and prevent a second submission.
- After durable task creation, show the task even while launch or delivery is pending. Display “Starting assistant,” “Instruction sent,” “Start not confirmed,” or “Delivery not confirmed” from the actual operation record. A lost response must not return the user to an empty form that invites a duplicate launch.
- Close/back preserves the draft under its destination without sending it. If a draft is persisted across app restarts, use protected storage; do not put goal text into route parameters, logs or notifications. Attachment handles can expire and must be revalidated when reopening.

The terminal composer keeps its current default meaning: opening an assignment picker without choosing a target does not reroute ordinary terminal input. The task composer is a separate, explicitly entered context. Reusing an editor component must not merge those two destination rules.

**Why:** the user supplies the goal instead of first designing an agent team. A visible destination and a single Start task button make the action predictable without an additional “Are you sure?” dialog for every normal request.

**UI-03 — Task detail: stable content, explicit sources, one primary action**

```text
‹ Tasks        Homepage visual update                More
website · Development Mac

Lead: Claude · Working
Last update 14:32

Needs your attention                                  1
Image service authorization                     [Review]

Plan from lead                                      View
Create artwork, adapt styling, then verify the page.

Work
Artwork             Codex       Working                  ›
Styling             Claude      Waiting for artwork       ›

Results                                                  ›
No result submitted yet

To: Lead · Claude                                  Change
[Add instructions…                              ] [Send]
```

- Section order is context → decisions → lead summary/plan → work → results. The composer is anchored above the keyboard/safe area. A transient toast may acknowledge a tap, but the durable operation status belongs in the task.
- A lead summary is labeled as agent-reported. Process observations carry observation time; stale status reads “Last seen working at 14:32” or “Status unavailable.” Do not change unknown into idle for visual neatness.
- With one assistant, show a compact assistant row and omit empty grouping chrome. With several workers, use vertically readable rows with title, agent, state and attention indicator. Avoid making a horizontal chip strip the only route to workers off screen.
- Tapping a worker opens a stable detail view; it does not select that worker as the recipient of the task's composer. Viewing and addressing are different actions.
- New output adds “New output available.” It must not move the reading position or replace the captured output. Refresh is explicit and applies to the selected view only. Returning from background/reconnecting preserves the pinned snapshot.
- A new decision increases the attention count and can announce one accessible notice. It must not steal focus, open a modal or switch terminals.
- Back returns to the previous list/selection/scroll state. Leaving the task does not stop work. The UI must not imply that keeping the phone awake is necessary for execution.

**Why:** the user can inspect one result or worker without losing the overall task. Attention is reserved for actual decisions, while unchanged content remains spatially stable.

**UI-04 — Follow-ups and recipient changes**

The default composer recipient is the current lead attempt, shown as “To: Lead · [name].” In worker detail, show “Message this worker” as an explicit action that opens a composer naming that worker. Returning to task detail restores the task's lead recipient.

Selecting a new recipient changes the destination chip and Send label before any request. Preserve text but revalidate attachments and scope. A disappeared/replaced assistant invalidates the selection; show “This assistant is no longer available. Choose a recipient.” Never substitute the replacement process or send to the ordinary terminal as a fallback.

If the lead is busy and the backend cannot accept a correlated follow-up, keep the message as a draft and show “Assistant is busy. Your draft has not been sent.” Do not pretend the message has entered a queue. A future inbox must be a separately implemented contract, with visible queued/canceled states.

During submission, freeze the addressed attempt for that operation and prevent double taps. A later UI selection cannot retarget an in-flight request. On success, clear only the submitted draft version; a newer draft typed during processing must survive. “Instruction sent” acknowledges delivery, not compliance with the instruction.

**Why:** looking at a worker should not accidentally redirect the next global instruction. Predictable addressing reduces memory burden and protects against the retained-selection defect found in App #77.

**UI-05 — Decisions and approvals**

Tapping Review opens a dedicated detail screen, with enough room to inspect the request:

```text
Decision required                                  Close
Task: Homepage visual update
Assistant: Codex · Artwork
Project: website · Development Mac

Requested action
[Actual operation/tool, affected paths or endpoint,
 and the original request/options as available]

Source: Native approval request / Terminal detection
Observed 14:34

[Deny]                         [Allow once]
Other available options                                  ›
```

The button labels above are illustrative: render only decisions the underlying request actually offers, preserving their real meaning. Do not invent “Allow once” when the provider only offers a broader grant. Keep original command/path/endpoint text inspectable and selectable; display it as untrusted content, not trusted App instructions. An AI summary is supplementary and never replaces the original request.

Rules:

- No option is selected or submitted by default. Reading, opening, scrolling or dismissing a request has no approval effect. No batch “Approve all.”
- On an explicit one-time decision, submit once against the displayed request identity and assistant instance. No generic confirmation dialog is needed after the user has already inspected and tapped the specific decision.
- A persistent grant such as “Always allow” opens an additional scope summary naming what permission persists and the actual revoke path when known. Confirm with a scope-specific label. If the scope/revocation behavior is unknown, say so; never imply that closing Muqun revokes it.
- Disable buttons during submission. If the request/instance changes, retain the old view as stale with “This request changed. Review the current request.” Do not silently replace the button meanings under the user's finger.
- If the outcome is uncertain, show “Approval not confirmed” and a Check status action. Do not resend the answer or state that it failed to reach the agent without evidence.
- If structured approval identity or safe answer delivery is unavailable, provide Open terminal to inspect the real prompt. Do not simulate safety by translating a generic Allow button into an unverified Enter key. Raw terminal interaction is an explicit user action, not a structured approval guarantee.
- Credentials and service login stay in the supported authentication flow. Do not ask the user to paste a secret into a task prompt as an approval workaround.

**Why:** meaningful friction belongs where authority changes. Reconfirming every normal send trains dismissal; a focused request review and stronger treatment of lasting grants concentrate attention on the actual consequence.

**UI-06 — Delivery failures, offline use and recovery**

| Situation | Visible state | Primary action | Required invariant |
| --- | --- | --- | --- |
| Definitive pre-send refusal | Instruction not sent + actionable reason | Fix the reason | No claim that a worker started |
| Launch created resources but a later step failed | Assistant needs attention + known terminal reference | Inspect assistant | Preserve partial resources; do not duplicate them |
| Acknowledgment lost | Delivery not confirmed | Check status | Never auto-resend |
| Offline with unsent text | Offline · draft not sent | Keep editing | Reconnection must not send automatically |
| Agent replaced/exited | Assistant unavailable | Inspect history / choose replacement | Old assignment is never rebound |
| Lead exits while workers remain | Lead unavailable · worker state retained | Choose a new lead | No automatic replay to workers |
| Result removed/changed | This result version is unavailable/changed | View submission details | No silent substitution of newer bytes |

“Send again” is secondary to Check status for ambiguous operations. It shows the original destination and the concrete warning “The earlier instruction may already have been received.” The explicit retry creates a new operation. Dismissing recovery UI does not resolve the uncertainty.

Local Mark reviewed/Remove from history remain available offline and never control an agent. A shared Accept result action requires server confirmation; while offline, offer local review only. Do not show a local click as shared acceptance.

**Why:** mobile connectivity fails at exactly the point where users are tempted to repeat a tap. The UI should expose uncertainty and preserve work, rather than make a timeout look like a safe retry.

**UI-07 — Results, revision review and previews**

A result detail shows the submitting assistant/attempt, submission time, summary, files/preview/diff, and verification evidence. Label evidence as agent-reported or independently observed. “Not tested” is a legitimate value.

Keep the chosen result version fixed. If a newer submission appears, show “A newer result is available” and let the user choose it. Acceptance records exactly the version displayed; it does not accept a result that arrived in the background. If the policy requires the latest version, a conflict asks the user to review it rather than silently redirecting acceptance.

Use “Request changes” to open a composer addressed to the lead, with the reviewed submission reference visibly included. Use “Accept this result” for the shared review action. Offer a local Mark reviewed separately where needed. Neither action merges, publishes, deletes files or terminates workers. Release/merge controls remain a separate explicit workflow.

Render Markdown and images through existing safe viewers. Agent-produced links and HTML are untrusted: no automatic external resource loading from result prose, no credentials in preview URLs, and no authenticated Gateway bridge exposed to generated web content. An external preview shows its host before opening. Never execute a file because an agent called it a preview. Sharing an artifact uses an explicit share action for the selected artifact/version.

**Why:** a readable result is the outcome users came for. Version binding makes approval meaningful and avoids accepting different content from what was inspected.

**UI-08 — Interruption, delegation limits and destructive actions**

Use specific actions rather than a single ambiguous Stop button:

- **Interrupt this assistant:** available in its detail, naming the assistant. One explicit action sends one interrupt request, provided identity is current. Show “Interrupt requested” until evidence supports a stronger state. It does not promise rollback or cancellation of all child processes.
- **Pause new delegation:** a task-level control, available only if Gateway enforces it. Explain that existing workers continue. It changes the policy for future managed dispatches, not execution already in progress. The explanation must say that it covers assistants started through Muqun; agents with independent shell/Herdr access may still start work outside that control. Do not present it as an OS-level stop or sandbox.
- **Hide from my history:** local, reversible where supported, no agent control. Offer Undo for local hiding rather than a routine warning dialog.
- **Delete a worktree / close all workers:** not a first-release shortcut. A future implementation needs a concrete resource list and unsaved/unmerged-work handling; it must not be bundled with acceptance or history removal.

Keep destructive or scope-expanding actions out of swipe shortcuts and away from frequent Send/Refresh targets. Every action must also be available through a labeled control, not a gesture alone.

**Why:** users need to know whether they are stopping input, preventing new delegation, or destroying resources. One generic Stop cannot honestly describe all three.

**UI-09 — Privacy, accessibility and interaction quality**

- First-release task notifications open the relevant task/decision; they do not approve or retry directly from the lock screen. Default notification text contains no prompt, command, path or result body. Respect existing app-lock/background privacy behavior on deep links and app switching.
- A deep link resolves IDs under the current paired identity, rechecks authorization and refreshes the live request before enabling mutations. A link to another machine must identify that destination; it must not inherit an armed composer action.
- Keep body text and actionable labels compatible with font scaling. Use text/icon distinctions alongside color for working, attention, stale and selected states. Screen readers announce the task, recipient, action and disabled reason; do not announce every terminal update.
- Use at least 44 × 44 pt touch targets on iOS and 48 × 48 dp on Android, accounting for actual hit areas and spacing. These targets follow [Apple's UI guidance](https://developer.apple.com/design/tips/) and [Android accessibility guidance](https://developer.android.com/guide/topics/ui/accessibility/views/apps-views). Smaller visual icons can sit within those targets.
- Restore focus to the initiating control after closing a sheet. Do not autofocus an approval choice. Support keyboard dismissal, hardware keyboards and reduced motion. Preserve important context at large text sizes rather than truncating the recipient or approval scope irretrievably.
- Busy states describe actual operations. Avoid indefinite success-colored spinners, decorative task percentages, auto-scrolling and animation that makes rows difficult to tap.

**UI-10 — Implementation mapping and acceptance script**

Use the existing APP-C1/C2/C3 and APP-T1/T2 packages. Add small task-list/detail, decision-detail and result-detail components around the shared composer/viewers; keep destination/operation state in testable hooks/reducers. Do not grow `server-terminal-workspace.tsx` into the task engine. Gate the new task surfaces on the new contracts, while preserving current terminal use.

Stable native test IDs should identify intent, for example `task-create`, `task-destination`, `task-start`, `task-recipient`, `task-send`, `task-decision-open`, `task-decision-submit`, `task-operation-check`, `task-result-accept`, and `task-history`. Dynamic rows must use stable task/attempt IDs, never list positions.

| Scenario | Pass condition |
| --- | --- |
| Start a simple task | A user can identify destination and chosen lead before one Start action; only one launch occurs |
| Change project with attachments | Text is preserved, destination/attachments are revalidated, no stale upload path is sent |
| Inspect a worker then send from task detail | Message still addresses the visibly selected lead; viewing did not retarget it |
| Double tap / reconnect after send | No duplicate mutation; receipt is queried and uncertainty is shown accurately |
| Switch Herdr → tmux with selection armed | No tracked assignment is sent; ordinary terminal use still works |
| Approval changes while open | Old action is refused with zero input to the new request |
| Two devices answer the same request | At most one matching native decision is executed; the other sees resolved/conflict |
| A new result arrives while reviewing | Displayed content and acceptance target remain the chosen version |
| Open app offline after an agent exits | History remains readable/manageable; no automatic stop or resend |
| New output while reading or using a screen reader | Position and focus remain stable; only an update indicator changes |
| Interrupt one worker | Only that verified worker receives the action; siblings remain unaffected |
| Malicious result content / notification link | No command execution, credential exposure or implicit authorization |

Run the native full suite and real paired checks specified by the repositories. Also conduct a small moderated usability pass on a compact phone and a large-text configuration: ask users to create a task, find a decision, identify the next message's recipient, recover from an unconfirmed delivery, and distinguish Accept from Publish. Record observed mistakes and assistance needed. The design rationale should be revised if those tasks expose confusion; this comment does not claim that fewer taps alone proves the workflow is better.

ryuhzk:
After #77 and osuki-dev/muqun-gateway#26 merged, the consolidated implementation handoff is now in documentation-only draft PRs #81 and osuki-dev/muqun-gateway#27. They include the detailed UI interaction/rationale, architecture, safety contracts, current implementation findings, dependency-ordered packages and acceptance tests. This PR remains the original research/design discussion; the new PRs provide reviewable repository documents for implementation after design confirmation.

PR #77, opened 2026-09-14
fix(collaboration): gate assignment on what the Gateway announces, not a constant

ryuhzk:
Pairs with **osuki-dev/muqun-gateway#26** — read that one first; this PR is the
other half of the same contract.

## Before

Whether a task could be assigned was decided by two things, neither of which was
the Gateway's answer.

**`AGENT_SPAWN_SHIPPED`.** `gatewaySupportsAgentSpawn` ANDed the `agent_spawn`
capability with a constant pinned to `false` since 1.2.0, so it returned `false`
for every gateway on every machine. "Start a new assistant" was unreachable in a
shipped build — and the demo session's one agent carries no `instance_id`, so the
existing-agent target was not offered either. The composer's assignment strip
therefore never appeared at all, and `e2e/.../agent-collaboration.ad` had written
that absence down as the expected behaviour.

**`collaborationAvailability`.** It has always known how to answer this properly
— the `agent_collaboration` capability plus a connected Herdr 0.9.0+ for *this*
session, with a different reason (`backend` / `gateway` / `herdr` /
`unavailable`) for each way of saying no. It had **no call site outside its own
test**. Nothing rendered those reasons, so the actionable upgrade explanation
AGENTS.md asks for did not exist in the UI.

## After

The constant is gone. The capability is the switch, and the strip is gated on
`collaborationAvailability` through one new pure function, `assignmentStripState`:

| availability | strip |
| --- | --- |
| `ready` | the targets, if there are any |
| `gateway` | toggle opens on one sentence: update Muqun Gateway and restart it |
| `herdr` | toggle opens on one sentence: update Herdr to 0.9.0+ and restart the session |
| `backend` (tmux) | no toggle, no sentence |
| `unavailable` | no toggle, no sentence |

tmux gets nothing on purpose: it has no agent instance identity to bind an
assignment to, there is no upgrade to recommend, and "upgrade" is not advice
anyone can act on there. Its composer is the ordinary composer it has always been
— **unchanged in every respect**, which is the rule AGENTS.md states for it.

Two orderings inside `collaborationAvailability` changed, and both are about
naming the right upgrade:

- **The session's backend is settled before the gateway-wide capability list.**
  Gateways now withhold `agent_collaboration` when no session has a Herdr 0.9.0+
  behind it, so reading that list first answered `gateway` for a machine whose
  Gateway is current and whose *Herdr* is old — the same "update the thing that
  is already up to date" this feature was fixed for once already (96e65dd).
- **`backends[].capabilities` is preferred over the version regex here** when the
  Gateway sends it: the Gateway pinged that backend itself. Its absence means a
  gateway older than the field, not a session that offers nothing, so the version
  fallback stays exactly as it was.

### On removing the kill switch rather than keeping one

No release note requires one — `release-notes/` holds only its README, and
nothing in the repo asks for a build-time hold-back. The comment's own reasoning
("a store debut is the wrong place to discover the next one") was about 1.2.0,
and what it actually bought since the collaboration composer landed was a feature
nobody could reach and a predicate that no longer answered a question about the
gateway. The gateway change beside this removes the reason a build-time switch
looked prudent at all: a capability answer is now evidence about the machine
rather than about the gateway's build date, so refusing on the capability is a
true statement in a way it was not before.

### Demo and e2e

`demoHealth()` now carries the per-session list
(`backends[0].capabilities: ['agent_collaboration']`) over its connected Herdr
0.9.0, so the offline flow exercises the route every current gateway takes rather
than the older-gateway fallback. `agent-collaboration.ad` asserts the offer
instead of the absence, and drives opening the strip, the "start a new …" chips
including the one the catalog reports as not on `PATH`, choosing a destination,
clearing it by tapping again, and closing.

The `new-task` flow stays disabled, for an honest reason this time:
`flows/new-task.ad` is an unconverted stub whose every step targets `${TARGET}`
with nothing to supply it. It must be rewritten against real selectors before it
can carry the `full` tag. (`suite.json` and the e2e README say so; the old text
cited `AGENT_SPAWN_SHIPPED`, which no longer exists.)

Three new UI strings, translated into all eleven languages.

## Compatibility

- **New app + old gateway.** Unchanged. No `backends[].capabilities`, so the
  top-level list plus `kind`/`version` decide, exactly as before. An absent key
  is treated as "older gateway", never as "offers nothing" — there is a test for
  that.
- **New app + new gateway.** The per-session list decides. A connected Herdr the
  gateway withheld the capability for can only be a version problem, so the app
  says *Herdr*, not *Gateway*.
- **New app + tmux.** No strip, no explanation, ordinary terminal use.
- **Old app + new gateway** is covered in the gateway PR.

## Checks

| check | result |
| --- | --- |
| `npx tsc --noEmit` | pass |
| `bun run lint` | pass |
| `bun run format:check` | pass |
| `bun test src` | **3700 pass, 2 skip, 0 fail** (176 files) |
| `bash scripts/e2e.sh --check` | pass — native syntax and manifest valid, 25 flows; runner tests 38 pass |
| `bash scripts/e2e.sh` | **not run — blocked, see below** |

### The Gateway half of the contract is proven against a real Herdr 0.9.0

The per-session branch of `collaborationAvailability` — the one that reads
`backends[].capabilities` and is preferred over the version regex here — depends
on the Gateway actually sending that key with the right contents. That was run
live against **real Herdr 0.9.0** (`herdr --version` → `herdr 0.9.0`, socket
protocol 22) on an isolated named session, with the gateway built from
`osuki-dev/muqun-gateway#26`:

```
top-level agent_collaboration: True
top-level agent_spawn:         True

default   tmux   connected=True  protocol=None  version=3.7c    capabilities=[]
herdr     herdr  connected=True  protocol=22    version=0.9.0   capabilities=['agent_collaboration']
stub090   herdr  connected=True  protocol=17    version=0.9.0   capabilities=['agent_collaboration']
stub089   herdr  connected=True  protocol=17    version=0.8.9   capabilities=[]
```

`herdr` there is the real 0.9.0 session, not a stub. A tmux-only gateway answered
`top-level agent_collaboration: False` with `agent_spawn` still present, which is
what makes `assignmentStripState` return `{ toggle: false }` on tmux rather than
an empty strip. No device was paired and no agent was started.

### The device suite is blocked on this machine, not skipped

It was written and validated on a Linux box with neither prerequisite AGENTS.md
names:

- the QA AVD `muqun_collaboration_qa` does not exist here (`emulator
  -list-avds` lists only the user's own `omarchy`), and AGENTS.md forbids both
  using the user's own device and creating an AVD per feature;
- `agent-device` here is **0.20.3**; AGENTS.md pins **0.20.10**;
- there is no iOS simulator on Linux at all.

So the full suite has to be run on the Mac, on
`muqun_collaboration_qa` / `muqun-collaboration-tests`, before this lands. The
rewritten `agent-collaboration.ad` is unverified against a device and is the
first thing to check.

## Not verified — outstanding on the Mac

**The paired App → Gateway → agent delivery check AGENTS.md requires.** Offline
fixtures cannot prove delivery, and nothing here claims they do. Against an
isolated Herdr session and a dedicated test device, with someone present to
answer trust prompts, and `proxy_on` in a new terminal:

1. **Herdr 0.9.0+** — the strip offers targets (the capability half of this is now
   proven live, above; what is not is that a task arrives): existing-agent
   assignment, new-agent startup, follow-up delivery, the actual returned output,
   and status changes without automatic terminal navigation.
2. **Herdr below 0.9.0** — ordinary terminal use unaffected, and the strip says
   *Herdr* is what needs updating (this is the path that used to name the wrong
   upgrade).
3. **A Gateway without the capability** — the strip says *Gateway*.
4. **tmux** — no toggle, no sentence, composer unchanged.

Then the full `bash scripts/e2e.sh` on each supported platform.


ryuhzk:
**Current collaboration review and implementation handoff**

Reviewed App head `78abddad87becaa32f3878138e6152a4e0fd505a` against base `00a8728a1a5ffd6d3238499038b179ccccf98d69`, together with Gateway PR #26 at `b6bca76`. This supplements the [module proposal on App #63](https://github.com/osuki-dev/muqun-app/pull/63#issuecomment-5659013436). Comments only; no repository files changed.

**Standards review**

**Existing defect — historical assignments are persisted but inaccessible.** `AGENTS.md` requires: “Keep earlier assignments as history; marking or removing history must work offline and must not stop an agent.” [`CollaborationNotice`](https://github.com/osuki-dev/muqun-app/blob/78abddad87becaa32f3878138e6152a4e0fd505a/src/components/collaboration-notice.tsx#L40) discards `history`, chooses `current[0]`, and returns nothing if there is no current task. Superseded/reviewed assignments and assignments whose agents departed become history. There is no other production UI consumer providing access to these records. Cold offline startup without live agents has the same problem. The local store actions themselves correctly avoid sending commands; the missing piece is an accessible history surface independent of agent discovery. This predates the gate change and should be tracked as a current feature defect, not blamed on this PR's new predicate.

For Gateway, the capability calculation introduced in `main.rs:387–399` directly checks backend kind and Herdr version. Gateway's `CONTEXT.md` places backend-specific branching inside adapters. Move capability evidence behind the backend metadata/port when extracting the shared services; HTTP should aggregate it. This is an architecture ownership issue, not evidence that the reported capability values are incorrect. Details belong on Gateway #26.

**Spec review**

**P1 — retained assignment state bypasses the new capability gate.** The requirement is that collaboration needs `agent_collaboration` and a connected Herdr 0.9.0+ backend for the selected session.

Reproduction path from source:

1. In a supported Herdr session, open the assignment strip and choose a new assistant.
2. Switch to a tmux session in the same mounted workspace screen.
3. Type a new message and send it.

[`resetSessionState`](https://github.com/osuki-dev/muqun-app/blob/78abddad87becaa32f3878138e6152a4e0fd505a/src/components/server-terminal-workspace.tsx#L1125) clears the draft and attachments but not the hook's `open`/`target` state or cached agent kinds. [`useComposerAssignment`](https://github.com/osuki-dev/muqun-app/blob/78abddad87becaa32f3878138e6152a4e0fd505a/src/hooks/use-composer-assignment.ts#L70) keeps that state across context changes. The [bar](https://github.com/osuki-dev/muqun-app/blob/78abddad87becaa32f3878138e6152a4e0fd505a/src/components/server-terminal-workspace.tsx#L3903) renders from `assignment.open`, while [sending](https://github.com/osuki-dev/muqun-app/blob/78abddad87becaa32f3878138e6152a4e0fd505a/src/components/server-terminal-workspace.tsx#L3330) branches only on `assignment.active`. The new toggle is hidden, but the armed operation remains. `spawnBoundAgent` then uses the current session, and Gateway intentionally accepts ordinary tmux spawn requests. Quick-action entry at `workspace:2395–2407` also arms assignment without checking collaboration availability.

The availability helper's truth table can pass while this interaction still violates the contract. Fix both destination-scoped state and the actual dispatch boundary; hiding controls alone is insufficient. This was established by source tracing, not a device reproduction.

**Validation gap — the feature is not yet demonstrated end to end.** The PR bodies explicitly leave the full App device suite and real paired App → Gateway → agent check outstanding. Health/version tests prove reporting, not existing-agent assignment, initial prompt submission, follow-up delivery, returned output or preserved focus. Preserve that qualification until those checks actually pass.

**Relationship to the planned module**

Current collaboration is a manual assignment feature: choose an existing/new assistant, send an instruction, keep local dispatch history and inspect a terminal snapshot. The proposed task module adds Gateway-owned tasks/attempts, explicit child relationships, recoverable delivery records, result submissions and human acceptance. They belong to the same product area but are different layers. Reuse the composer, terminal viewers, adapters and delivery services; do not build a second transport or retrofit the new task meaning into the old local `CollaborationTask` type.

In particular, a durable task module is not required to fix the two current defects above. Conversely, a working assignment strip does not mean the task module is implemented.

**Executable work packages after the design is confirmed**

**APP-C1 — enforce collaboration at dispatch and scope selection.** Depends on the selected-session capability contract, not the future task service.

- Files: `src/hooks/use-composer-assignment.ts`, `src/components/server-terminal-workspace.tsx`, `src/lib/agent-collaboration.ts`, and quick-action assignment request handling/store as needed.
- Bind assignment selection to `{serverId, sessionId, sourcePaneId}`. Changing destination invalidates selection/command and pending selection requests. Preserve draft text where appropriate but require deliberate re-selection; never silently fall back from a blocked assignment to ordinary terminal input.
- Scope the cached catalogue to the Gateway and discard stale asynchronous catalogue results. A same-Gateway session change can reuse host facts, but must recalculate session eligibility.
- Check the selected session immediately before each tracked dispatch, after upload/read awaits, including shortcut entry and existing-agent sends. Keep ordinary tmux composer/spawn functionality separate.
- Regression cases: armed Herdr → tmux switch; Herdr → another server; capability loss while composing/uploading; quick action into an unsupported session; stale catalogue response. Assert zero collaboration mutation requests on refusal, and that draft/attachments are not sent to the ordinary terminal instead.
- Done when these cases pass and the supported path still delivers with no implicit navigation.

**APP-C2 — expose all current assignments and offline history.** Can follow C1 independently of Gateway task storage.

- Files: `src/components/collaboration-notice.tsx`, `src/hooks/use-collaboration-output.ts`, `src/stores/agent-collaboration.ts`, and a focused history/list component.
- Provide explicit selection among current assignments rather than only `current[0]`. Render reviewed/superseded/departed assignments from stored history, including when disconnected.
- Local review/removal must send no Gateway request and must not stop an agent. Loss of identity disables terminal actions but does not delete history. New output updates an indicator; it does not replace a pinned snapshot. Reconnection alone must not act as a user refresh.
- Regression cases: two simultaneous assignments; review one then remove it from history; agent exits; restart offline; reconnect with a replacement occupant; receive output while reading an older snapshot.
- Done when all histories remain manageable without live agents and terminal actions require a current verified target.

**APP-C3 — truthful outcome copy and recovery.** Depends on C1.

- The current send path uses “Assistant started” for an existing-agent assignment and “A terminal was created” for its unconfirmed delivery. Branch outcome copy by operation: “Instruction sent” for an existing assistant; only mention creation when the response confirms it.
- Preserve known pane/instance and partial-step information in a recovery view. Distinguish explicit refusal from unconfirmed effects. Offer deliberate inspection/retry; never auto-retry or imply that a timeout proves failure.
- Regression cases: existing-agent success/refusal/lost acknowledgment; successful new launch; launch with unconfirmed initial prompt. Validate both displayed wording and number of mutation calls.

**APP-T1 — task-first single-assistant UI.** Depends on Gateway durable task/query/delivery/result contracts described on #26; keep this out of the compatibility fix.

- Reuse the composer for project + goal + attachments, with Automatic as the default and explicit assistant choice as an advanced option.
- Add task list/detail, operation outcome, registered results and review actions. Keep terminal observation as a drill-down. Normal follow-ups address the current lead attempt; worker-directed instructions require explicit selection.
- Use Gateway task IDs/revisions for shared state. Keep local read markers and legacy assignment history separate. Capability absence leaves existing terminal/assignment flows usable.
- Done when create → initial response → follow-up → reconnect → submitted result → request changes/accept works for one assistant before delegation UI is introduced.

**APP-T2 — delegated work view.** Depends on APP-T1 and Gateway's scoped child-task operations.

- Render explicit parent/child relations and recorded dependencies, never branch-name prefixes. Bound concurrent views; read output only for the selected assistant.
- Distinguish observed process state, agent-reported plan state, submitted results and user acceptance. Lead loss must not hide workers or prior submissions. No invented completion counts or percentages.

Each implementation package must add relevant native `.ad` coverage with the `full` tag and complete the five App gates. Startup/delivery packages also need the real paired check on an isolated session and dedicated device. Test both supported platforms before claiming both. This comment is an implementation handoff, not authorization to skip those gates or publish a release.

**Review evidence:** the existing collaboration/spawn unit tests on local main `be4293e` passed: 35 tests, 0 failures, 107 assertions. This is explicitly not the PR-head device suite; it does not exercise the retained-state interaction above. No app/device mutation was performed.

Standards: one current App history requirement violation, plus a Gateway ownership issue. Spec: one concrete capability-bypass defect and outstanding required validation. The durable task design is future scope, not a retroactive missing requirement for this PR.

ryuhzk:
**UI/safety acceptance criteria for the collaboration fixes**

The new [UI interaction specification on #63](https://github.com/osuki-dev/muqun-app/pull/63#issuecomment-5659175202) defines the task/worker/decision/result screens, action labels, recovery states, accessibility expectations and the reasoning behind each interaction. The [Gateway contract follow-up](https://github.com/osuki-dev/muqun-gateway/pull/26#issuecomment-5659184482) defines the server guarantees those screens require. Both are comment-only proposals.

For this PR's current collaboration scope, apply these concrete criteria without pulling the entire future task module into the fix:

- **Addressing:** the chosen assistant and destination remain visible before Send. Switching server/session invalidates the armed assignment. Inspecting another worker must not silently retarget a composer. Capability failure never falls back to sending the same text as ordinary terminal input.
- **Entry paths:** apply the same eligibility and destination checks to the toggle, shortcuts, resumed drafts and the actual send handler. Assert zero tracked-dispatch requests after a switch to unsupported tmux/Herdr, including when attachments were uploading.
- **Outcome language:** existing-agent success says “Instruction sent”; only confirmed new launches say an assistant was started. A timeout says “Delivery not confirmed,” not “Nothing was sent.” Known partial resources have a usable inspection action. Retrying uncertain delivery is an explicit new action, never automatic.
- **History and reading:** reviewed, superseded and departed-agent assignments stay accessible offline. Local review/removal sends no command. New output and reconnection do not replace a pinned snapshot or move the viewport. Provide explicit current-assignment selection rather than requiring the user to remove newer cards to reach older ones.
- **Approval behavior:** do not replace the original request with an AI summary, answer on opening, preselect a decision or replay input after uncertainty. Persistent grants need their actual scope explained. Current `use-pane-approval.ts:127` reports “That answer did not reach the agent” for a generic failure; that wording is too strong for a lost response after possible delivery. Use an unconfirmed outcome and status inspection unless no delivery is proven.
- **Native approval integration:** Gateway returns native `approval_id`, while the current App parser requires a fingerprint. The future decision UI needs an explicit native/menu identity union and a strict request-bound answer contract; do not manufacture a fingerprint simply to make the existing parser accept native requests. The Gateway comment records the corresponding source-level stale-request gap and its limitations.
- **Interaction quality:** preserve text at large font sizes, recipient context, screen-reader focus and usable touch targets. Status must be available in words, not just color. Notifications may open a view, but must not approve, retry or change an armed recipient.

These are observable behaviors for APP-C1/C2/C3 and the later APP-S1/S2 packages. The detailed specification includes scenario-based acceptance tests and a small usability evaluation: users should be able to identify the recipient, recover from uncertainty, and distinguish acceptance from publishing without assistance. That evaluation has not yet been run; the rationale is a hypothesis to test, not a claim of measured usability.

ryuhzk:
The post-merge design and implementation handoff now continues in #81, paired with osuki-dev/muqun-gateway#27. These are documentation-only draft PRs covering current collaboration findings, task/UI design, safety contracts, implementation packages and acceptance criteria. The original review comments remain available here.

PR #81, opened 2026-09-14
docs: task workflow architecture, UI specification and implementation handoff

ryuhzk:
The task-workflow design and implementation review were spread across comments on #63 and the now-merged #77. This PR turns that discussion into a reviewable, English-only implementation handoff based on merged App `a178514` and Gateway `f48568c`.

- `docs/task-workflow-design.md` defines ownership, the relationship to current collaboration, remaining source-reviewed defects, compatibility rules, and dependency-ordered APP work packages.
- `docs/task-workflow-interactions.md` specifies screens, recipients, approvals, uncertain delivery, offline history, result versions, privacy, accessibility, rationale, and acceptance scenarios.

This is a draft for design confirmation. It introduces no runtime changes and does not claim the task module or its security guarantees are implemented. Relevant source files were compared with the earlier reviewed PR heads; the recorded current defects have not been silently treated as fixed by the merge.

The default direction is one lead assistant, explicit task/attempt identities, Gateway-owned durable facts, registered results, and agent-owned planning. Detailed backend contracts belong to the companion Gateway design PR. #63 remains the research/design discussion; this PR provides the consolidated execution handoff after #77 merged.

Validation on this branch:

- `npx tsc --noEmit` — passed.
- `bun run lint` — passed.
- `bun run format:check` — passed.
- `bun test src` — 3743 passed, 2 skipped, 0 failed.
- Documentation file links, code-fence pairing, English content and `git diff --check` — checked.
- Full device E2E and paired delivery checks — not run for this documentation-only change. They remain mandatory acceptance work for implementation packages that touch app/startup/delivery code; no platform or real-delivery coverage is claimed here.

Follows #77; related design discussion: #63. No release/tag or agent execution behavior changes.

Companion Gateway design: https://github.com/osuki-dev/muqun-gateway/pull/27


ryuhzk:
Implementation architecture update: consolidate collaboration and managed work

The implementation will extend the existing collaboration feature. It will not add a second assistant dispatcher, composer, or terminal-output surface.

| Responsibility | Owner and reuse decision |
| --- | --- |
| Select an assistant, send a follow-up, inspect terminal output | Existing collaboration controls and shared delivery transport |
| Task intent, execution attempts, parent/worker relationships, durable operation outcomes | Gateway work service; App renders its explicit records |
| Local dispatch history | Existing store remains available offline; it is not silently converted into server task history |
| Result acceptance | Explicit review of an immutable submission ID and displayed task revision; idle status is not completion |
| New task entry | Evolve the existing New Task route and reuse the shared composer; preserve legacy startup on older Gateways |

Local implementation now includes session-qualified assignment ownership, stale-catalogue protection, revalidation across asynchronous delivery boundaries, truthful startup/send outcomes, explicit history selection, and pinned output snapshots. Selecting a newer assignment or reconnecting must not replace the output currently being read. Removing/reviewing local history must not send an agent command.

The controlled task detail component separates attempts, uncertain operations, submitted results, and human reviews. Selecting an old attempt does not retarget the composer. An uncertain operation offers a status query, not an automatic resend. Result review callbacks carry the exact submission ID and revision shown to the user. Artifact paths will not open mutable live files as if they were the reviewed version.

Security dependency discovered during implementation: Herdr 0.9.0 resolves prompts through a current pane/name, and names can be renamed and reused. App preflight checks and a random launch alias cannot establish strict instance-bound delivery. Gateway must withhold that stronger capability until the native input path enforces the expected launch identity through submission, including delayed Enter. See the corresponding Gateway PR for the native protocol evidence and execution plan.

Validation status: focused collaboration checks pass locally; full integration, the full Android suite, and the real paired delivery check remain outstanding. Android QA uses an isolated secondary user on the existing Omarchy emulator, preserving the original user's pairing data. iOS runtime testing is explicitly deferred to the user and will not be claimed as locally verified.

Follow-up agents should implement in this order: finish shared contracts and persistence tests; integrate the existing route/composer with task records; implement and prove native instance/approval binding; connect version-verified artifacts and review; complete dependency/worker policy; run all repository gates and the isolated paired workflow. Keep this PR draft until the applicable gates pass. A storage-only or UI-only implementation is not completion of the requested workflow.

ryuhzk:
Real Android integration update

The isolated Omarchy QA user has now paired with the implementation Gateway using required application transport encryption. A command sent from the App, `echo MUQUN_SHELL_$((2700+29))`, produced `MUQUN_SHELL_2729` in both the real Herdr pane and the App terminal display. This proves the shell transport/output path, not AI completion.

The installed Codex 0.154.0 produced a useful real failure case: Herdr 0.9.0 reported `idle` and `interactive_ready: true` while Codex was displaying its directory-trust question. The App correctly displayed the approval choices. I did not answer that prompt; confirmation is pending under AGENTS.md. Actual AI response, follow-up and App-created assistant success tests remain outstanding.

A fixture captured from that real screen now tests the shared Gateway prompt path. The path refuses a detected approval before any `agent.prompt` call, and the legacy delayed-Enter path also refuses detected approval menus. Existing-agent delivery and startup prompt delivery share this check. It is a conservative refusal check, not proof of atomic instance binding or detection of every possible dialog.

The App's collaboration regression flow passed its demo startup/reviewed-history/removal assertions. The complete Android suite has not passed: the offline theme-catalogue assertion encountered a live online catalogue; the run was subsequently interrupted for additional lifetime/cache fixes. The next complete run must establish the required offline conditions while preserving Metro access. No ready status or iOS runtime coverage is claimed, and local implementation has not yet been pushed as a completed feature.

ryuhzk:
### Implementation update: native delivery identity and validation

The isolated Android App-to-Gateway-to-Herdr shell check returned `MUQUN_SHELL_2729` from `echo MUQUN_SHELL_$((2700+29))`, confirmed in both the App and native terminal output. This is transport evidence, not AI completion evidence. Codex remains at the isolated directory trust prompt; no trust response has been automated. Real AI startup, initial instruction, follow-up and returned-answer checks remain pending that approval.

Source inspection of official Herdr v0.9.0 found that legacy agent startup writes into a persistent shell. A pane or terminal identifier cannot by itself identify the assistant lifetime: the shell can receive input after the assistant exits. Native work now targets a dedicated agent-owned PTY, immutable launch identity, and identity validation at both queued text and submit-key writes. Native approval identity is a separate guarantee; terminal pattern matching does not establish it.

Gateway has a typed bound-delivery port whose shipped adapters currently refuse unsupported operations. No strict execution capability is advertised. The durable execution service reuses existing placement/start primitives, records confirmed partial resources, distinguishes startup from prompt delivery, and does not replay uncertain operations. App integration retains existing collaboration and adds a task context with explicit recipients and separate operation receipts.

Artifact storage now enforces 1 GiB and 16,384 aggregate entries, including interrupted captures and orphaned files, with cross-process quota locking and verified digest reuse. Pausing delegation blocks new descendant launches while preserving existing follow-ups; resuming changes policy without automatically starting queued work.

Interim App validation passed TypeScript, lint, formatting and 3,774 tests (2 skipped). The full Android suite has not passed yet; the previously documented offline catalogue setup issue remains to be corrected before the next full run. iOS runtime coverage is unverified. These PRs remain draft; ongoing native and App integration still require their own checks and the real paired gate before readiness can be claimed.

ryuhzk:
Implementation and validation update — not ready for merge.

The real paired Android shell check has passed: App → encrypted Gateway transport → isolated Herdr shell → returned `MUQUN_SHELL_2729`. This is transport evidence, not an AI conversation. The isolated Codex process still awaits the directory-trust decision; no trust prompt was answered. Real AI startup, initial instruction, follow-up, and returned-output checks remain pending.

Current local checks: App TypeScript, strict lint, formatting and 3,837 unit tests pass (two existing skips). Gateway formatting, all-target strict Clippy, and 605 Rust tests pass (20 ignored). The native Herdr implementation passed its prescribed lint and test gates, including 3,281 Rust tests; its isolated process smoke used an explicitly simulated assistant and verified UTF-8 delivery, wrong-generation refusal, and refusal after exit. These counts describe the checked snapshot before the new lifecycle work below.

Android testing found and drove fixes for missing delivery receipts and task-sheet dismissal during scrolling. Tasks now uses full-screen presentation with explicit Close. The next run accepted the exact first result submission correctly, then exposed a row under the fixed header that could not receive the next tap; the content boundary is being corrected. Existing exact submission IDs and semantic review assertions are retained. The full Android suite has not passed; iOS runtime testing is unverified.

A release-blocking recovery gap was identified: a lead reservation can survive forever after an assistant exits or a definitive startup refusal leaves workspace resources. The implementation plan is now recorded in Gateway `docs/task-workflow-lifecycle.md`:

1. Separate durable reservation lifecycle from operation outcomes and retained workspace/history records.
2. Commit a dispatch fence before native startup. Release a never-dispatched attempt only while atomically preventing a suspended continuation from starting it.
3. Add native ownership epochs and an exact-generation lifecycle query. Only successful owned-child wait/reap proves exit. Wait failures, missing aliases, native restart, evicted evidence, and lost launch identity remain unknown and reserved.
4. Add an explicit paired-device reconciliation action with idempotent receipts, expected revision/identity, atomic reservation/event updates, and local authority revocation. GET receipt recovery never performs reconciliation or retries a mutation.
5. Offer “Check assistant lifecycle,” then a separate “Start replacement” only after verified release. Preserve previous results, output snapshots, and resources; do not stop assistants, delete workspaces, or label the task complete.

Native evidence and Gateway store/service implementation are in progress. This extends the shared collaboration execution foundation rather than adding a second launcher. Android process-kill recovery and the real paired AI gate remain required before readiness can be claimed. No release, merge, signing change, or installed Herdr replacement has been performed.

ryuhzk:
Validated milestone; implementation remains draft.

The full Android suite completed with **17/18 flows passing**, including manual collaboration and the complete managed-task flow. The failing theme flow checked reset visibility before scrolling to the control. That sequence has been corrected, and its reset assertion now checks the terminal opacity value specifically rather than accepting an unrelated `100%` label. Native syntax/runner checks pass; the final full suite must be repeated after the remaining App changes.

A fresh, authenticated Gateway–Herdr process integration passed:

- Create a task and retain an acknowledged immutable launch identity.
- Send the initial instruction as its own operation, then a follow-up, and observe actual returned fixture output.
- Let the owned child call the actual Gateway CLI using its injected local context to submit a result and a 29-byte artifact.
- Retrieve the original immutable artifact even after changing the source file.
- Observe the exact owned generation exit, explicitly reconcile it, reject the old grant, and start a replacement with a new attempt and launch identity.

This used a deliberately simulated native assistant and loopback paired-device authority. It is **not** encrypted App-to-agent or real AI evidence. Both private processes stopped cleanly. The earlier encrypted Android shell echo remains valid; the real Codex conversation still awaits the requested directory-trust decision.

The process check caught a defect unit tests had missed: a created-only native receipt legitimately lacks an observed agent type. Gateway incorrectly rejected it after the process had already started. The fix adds a declared startup profile to the receipt, validates it separately, preserves the launch identity, and keeps readiness unknown until runtime observation establishes it. Conflicting declared/observed types remain rejected. The checked Gateway snapshot passed 32 library and 591 binary tests (20 ignored); native checks passed 3,286 Rust tests plus supporting suites and Linux/Windows lint. Later input work is a separate, still-in-progress snapshot.

The remaining first-loop implementation is explicit:

1. App lifecycle checks and separate replacement actions, with durable request recovery and no automatic resend.
2. Scoped input attachments and recent-project selection. `work_inputs_v1` uses paired session/project upload receipts, 10 MiB per file, at most nine references, the existing shared immutable-store quota, and transactional task/operation claims. Text-only clients remain compatible; arbitrary caller paths are not accepted as input authority.
3. “Request changes” records the exact submission review, then prepares a visibly lead-addressed draft containing that submission reference. It never sends automatically or silently replaces an existing draft.
4. Actual Android process termination after a committed request loses its acknowledgment, followed by receipt-only recovery with no duplicate POST.
5. Real paired AI startup, instruction, follow-up, output/result evidence, then the final full Android gate. iOS runtime remains user-deferred and unverified.

The working-tree handoffs are `docs/task-workflow-lifecycle.md`, `docs/task-workflow-inputs.md`, and App `docs/task-workflow-completion-map.md`. Delegation/dependencies, strict decisions, and tablet/accessibility work remain visible authorized scope; managed tmux and cross-machine/destructive operations retain their explicitly later placement. No merge, release, signing change, or installed Herdr replacement has occurred.

ryuhzk:
Android process-restart recovery has now been exercised against an isolated, encrypted Gateway using the real App, native networking, and SecureStore. This is additional device evidence, not an AI completion claim.

- Manually paired the dedicated QA Android profile through the normal code-pairing flow.
- Created one task and deliberately disconnected the proxy after Gateway committed the successful response, before Android received it.
- The App displayed **Operation not confirmed** and disabled another submission.
- Force-stopped and relaunched the App. Opening Tasks recovered the original task through `GET /work/receipts` and task reads. It did not start an assistant or resend a mutation after restart.
- Read-only SQLite verification found exactly one task, one create receipt, and zero attempts. The recovered screen explicitly stated that no assistant had been started.

The fault injection also exposed a release blocker: before restart, the native transport automatically replayed the original POST after the disconnected acknowledgement. Gateway rejected the second request with HTTP 409, preserving one stored task, but an implementation claiming no automatic mutation retry must address the networking layer as well as JavaScript. Source tracing identifies the Android Nitro Fetch/Cronet upload rewind path as the likely source. A reproducible native fix and a rebuilt-APK rerun are required before this test can pass the strict one-POST criterion. The independent Gateway receipt, replay protection, and dispatch fences remain necessary defenses.

Current App static/unit gates pass: TypeScript, lint, formatting, and 3,856 tests (two skipped). The last full Android suite was 17/18; the custom-theme test correction and expanded lifecycle flow still need the next full run. Attachment integration is proceeding with exact draft/recipient capture, read-only uncertain-upload recovery, and explicit expired-input re-upload. Request changes must prepare an addressed draft only after the exact review is saved; it must not send automatically or overwrite newer text.

The real isolated shell echo succeeded earlier. The actual Codex AI test still awaits directory-trust approval under the repository's explicit rule; simulated process tests do not substitute for that gate. iOS runtime testing remains deferred to the owner. Implementation is still local and these PRs remain drafts.

ryuhzk:
The next implementation packages retain the existing task/collaboration foundation and define explicit UI authority boundaries. The detailed Gateway service/store contracts are recorded in the companion Gateway PR. These are implementation requirements, not completed device coverage.

**Current first-loop changes:** scoped attachment composition and exact result-review follow-ups are implemented locally. Attachments capture the project, draft, recipient, and exact file/annotation versions before awaiting uploads. A successful older send clears only its submitted versions; later edits survive. Display-name changes no longer erase staged task attachments, while pairing or destination changes invalidate their authority. Request changes saves the exact submission review, then prepares a lead-addressed draft without automatically sending or overwriting existing text. All 1,036 current messages per locale are translated; demo additions will require another catalog check.

**Delegation UI and rationale:**

1. Keep one lead as the initial task experience. An additional-worker limit is a capacity limit, not permission for unlimited child creation or automatic approvals. Add a separate explicit delegation setting, default disabled for old tasks, showing allowed profiles, project, lifetime child limit, concurrency, and dependency requirement.
2. Show child tasks as stable vertical rows within Work. A row opens that child's history; it does not redirect the parent composer. Keep the recipient label explicit. Child startup and initial delivery remain separate recorded operations, even if the lead requests them in sequence.
3. Display dependency readiness as facts: waiting for a result, waiting for acceptance, paused, unavailable authority, or capacity unavailable. Idle or process exit never satisfies a dependency. Show the exact pinned submission; do not silently replace it with the latest result. A newly ready dependency does not automatically launch a task.
4. Pause blocks future managed dispatch at the documented policy boundary. It does not stop existing assistants or undo a dispatch already claimed. Keep follow-ups and result review available.
5. When the lead disappears, preserve child results, worktrees, history, and reservations. Show coordinator unavailable. Replacement and delegation takeover are separate explicit actions against exact attempts/revisions; do not replay the predecessor's uncertain commands. Workers retain their own result-reporting authority.

**Decision and interruption UI:**

- Current adapters do not establish immutable approval occurrence identity. Until a provider does, render inspection and Open terminal, with an actionable unavailable explanation; never turn an old fingerprint into a structured Allow button.
- A future strict decision view must pin exact provider request/options and assistant identity. No option is preselected. Persistent grants show their actual scope and revoke information before a scope-specific confirmation. A changed request stays visibly stale instead of changing button meaning beneath a touch.
- Exact-instance interruption is separate from permission decisions. Show Interrupt requested only from an actual write receipt, never Stopped or Complete. It must not release a worker slot or revoke result authority just because a cancellation key was sent. Uncertain interruption has read-only status recovery and no automatic resend.

**Tablet and accessibility acceptance:** use the same selected task, pinned result/output, recipient, and drafts in phone and tablet layouts. Tablet presents a stable list beside detail; resizing must preserve selection/scroll and send nothing. Keep composer actions above the keyboard/safe area, at least the shared touch-target size, with distinct accessible names for inspection, addressing, and mutation. Verify large text, compact width, reduced motion, focus order, and explicit update notices. Do not infer this coverage from desktop screenshots or unit tests.

**Required validation before readiness:** all five App gates after final source changes; native full-tag flows for attachments, exact review draft, lifecycle/replacement, and later delegation/decision surfaces; real paired initial/follow-up/file/output/result checks; rebuilt Android lost-ACK test asserting one POST plus read-only recovery. The native patch compiles and a debug APK exists, but runtime proof awaits a dedicated QA AVD: only Omarchy exists, and installing there would update both Android users' shared binary. Creating the absent QA AVD requires explicit permission under AGENTS.md. The real Codex trust prompt is separately awaiting the required approval. iOS runtime remains owner-deferred and is not claimed.

ryuhzk:
# Managed task UI completion proposal

Status: implementation proposal only. App source and device automation remain frozen while the parent runs the full Android suite. This document does not claim implementation or platform coverage.

## Findings from the current implementation

- `src/components/work-task-workspace.tsx` uses one full-screen `SettingsSheet`, which owns one vertical `ScrollScreen` capped at 640 points. List, detail, and creation are mutually exclusive render branches. There is no task-specific tablet split layout or viewport restoration.
- `src/components/work-task-list.tsx` renders the title, project, selection, and delegation-paused flag from `WorkTask`. It cannot truthfully distinguish an unreviewed result, unresolved delivery, or last durable activity. Fetching each detail to invent these labels would introduce N+1 requests and would still be wrong when detail collections are paginated.
- `WorkController.showList()` currently discards detail and selected attempt/result/recipient IDs, preserving only the draft by task ID. Returning to a task loads a fresh detail. This breaks persistent historical selection across back navigation and would make responsive remounting unsafe.
- `use-work-output.ts` owns pinned output snapshots inside the mounted hook. A responsive layout must keep that hook mounted once. Re-entry persistence needs a deliberate bounded cache of its existing snapshot type, rather than another output observer or polling loop.
- `SettingsSheet.fullScreen` deliberately places a fixed safe-area viewport outside its scroll content. Real Android evidence previously showed that moving safe-area padding lets accessibility see untappable rows behind the status bar. Preserve that correction. Default native form sheets must keep the existing direct scroll-root contract.
- `TerminalComposer` already owns the shared input/send UI. The terminal workspace uses one `useReanimatedKeyboardAnimation()` source for its dock and viewport because an independent `KeyboardStickyView` previously became displaced. The task layout should reuse this principle and the existing composer, not layer two keyboard avoidance mechanisms.
- Gateway `work_http.rs::TaskPage` returns `{tasks,next_after_id}`; `store.rs::list_tasks` only reads task records in ID order. The records and changes tables already contain the facts needed for a read projection. The newer delegation/dependency models must remain the authority for dependency decisions.

## UI behavior and component boundaries

Introduce a task-specific `WorkTaskLayout`, using the existing sheet ground, chrome, theme tokens, and safe-area primitives. Do not widen or restructure all `SettingsSheet` consumers. The tasks route remains a full-screen route on both platforms.

Use the actual safe content width, not device type or orientation. Start with a 320-point list pane, 24-point divider/gutter, and at least 480 points for detail. Enter two panes only when these minimums fit; at enlarged text sizes, increase the required pane widths or fall back to the single-pane layout. Final breakpoint is a measured layout constant covered by tests, not a claim that all tablets must split.

Keep one controller, one output observer, one artifact-preview owner, and one set of attachment hooks above responsive layout branches. Width changes only affect placement. They must never invoke selectTask, showList, refresh, addressAttempt, or a mutation. Prefer a stable component tree with flex direction/display changes over remounting detail and composer at a breakpoint.

In two panes, the task list remains visible and independently scrollable. Selecting a row opens detail on the right without moving the list. Loading a different detail retains the current detail with a visible loading indicator until the new request succeeds; failure keeps the previous reading snapshot and draft. The selected row reflects the successfully displayed task. New task opens the existing creation form in the detail region, without silently discarding another task's draft. In one pane, Back returns to the list; returning to the previous task restores its selected historical attempt/result and pinned snapshot. Hardware Back first dismisses the keyboard, then returns from attachment detail/creation/task detail, then closes the route using normal navigation.

Separate detail's scrollable reading area from a bottom composer dock. The dock contains the explicit recipient identity, attachment strip, text field, and send control. It must remain inside the task safe viewport and rise from one keyboard animation source. Measure dock height and reserve matching bottom clearance for the reading viewport. Do not combine translated dock, KeyboardAvoidingView, and a second sticky keyboard controller. Collapsible long attachment details and review editing remain in the reading area. Limit composer height by available viewport, permitting internal input scrolling so 200% text plus a keyboard still leaves the recipient and send control reachable. Large hardware-keyboard windows use the same dock without an artificial keyboard gap.

Output remains refresh-pinned: SSE and polling update a notice only. Showing the dock, rotating, resizing, changing focus, or receiving a result must not replace output, change the selected result, or scroll to the end. Explicit output refresh replaces that snapshot and restores the nearest surviving anchor; it does not implicitly choose the newest result. A separate explicit Latest result action may be added later with its own acceptance tests.

## View memory and ownership

Add a bounded presentation-memory helper scoped by the existing complete pairing fingerprint and session. Store no new authorization or mutation state. It records list position, task detail positions, selected attempt/result/recipient IDs, and references to the already parsed reading snapshots. Journal hydration and pending-operation locks remain solely in the current controller/journal.

Represent viewport positions as a stable item/section ID plus relative offset, with a clamped raw offset fallback. List anchors use task IDs. Detail anchors use section plus immutable result/review/operation/attempt ID. Output anchors additionally include the existing pinned snapshot identity. Restore after layout/content measurement; cancel restoration if the user has started dragging. Do not continuously issue scroll commands from state updates. On large-text or width changes, remeasure the anchor rather than replaying an absolute pixel position.

Change navigation state explicitly: `view: list | detail | create` is presentation state and must not clear the active cached detail. Add a same-task return path that does not fetch or reset selection automatically; explicit Refresh remains the authority to replace records. Bound in-memory history by both task count and bytes (proposed 20 task snapshots / 16 MiB). Never evict the active or unresolved task; reject caching additional snapshots if necessary. Eviction may lose old viewing position but must never remove pending journals, draft ownership, or receipt identities. Process restart can restore lightweight view IDs/offsets if a separate validated local view preference is added; do not persist terminal output or result content incidentally. A fresh process loads records explicitly and keeps mutation recovery read-only. Pairing replacement/unpair clears its presentation cache; a safe label rename retains it.

## Gateway summary request

Add an optional `work_task_summaries_v1` capability and a compact endpoint:

`GET /api/sessions/{session_id}/work/task-summaries?limit=20&after_id=<uuid>&snapshot_cursor=<cursor>`

First page omits after_id and snapshot_cursor. Response:

```json
{
  "items": [{
    "task_id": "uuid",
    "session_id": "session",
    "task_revision": 12,
    "title": "Update homepage",
    "repo_path": "/project",
    "paused": false,
    "last_activity": {"cursor": 91, "kind": "result_submitted", "entity_id": "uuid"},
    "reserved_attempts": 1,
    "unresolved_native_operations": 0,
    "unreviewed_results": 1,
    "latest_result": {"submission_id": "uuid", "review": null}
  }],
  "snapshot_cursor": 91,
  "next_after_id": null
}
```

Make absent activity/latest result explicit null. A review in latest_result is `{review_id,decision}` and applies only to that submission. Include no brief, result summary text, artifact arrays, prompts, credentials, native output, or full record bodies. Counts are bounded safe integers. Unknown event kinds render a neutral Activity recorded label. Prefer no event time in v1: the existing changes table has no timestamp; task updated_at can be presented as task-record updated time but must not be mislabeled as precise event time.

Initially expose factual labels rather than a mutually exclusive task status: Operation needs checking, Result awaiting review, Latest result accepted, Changes requested for latest result, and New delegation paused. Accepted means a human reviewed that exact submission; it never means the entire task or agent has completed. Reserved attempts means capacity is retained, not that native processes are currently running. Refused historical operations should not permanently mark a task failed. Derive unresolved native operations using the existing lifecycle-aware start/delivery blocker semantics; exclude uncertain read-only reconciliations and released generations exactly as the current service does.

Do not add dependency status by reimplementing scheduler rules. Gateway owner confirmed `WorkStore::dependency_readiness(session,task)` already returns `Ready | WaitingForResult | WaitingForAcceptance` using exact `Task.dependencies` and parent delegation policy; latest review uses insertion order. Its current public method is a single-task connection read, so extract/reuse its transaction-scoped evaluator with batched prerequisite loads before exposing summary readiness. Prefer an optional `dependency_readiness` enum matching those existing facts over a newly invented blocked count. Omit it until that evaluator can provide a bounded projection in the same snapshot. App must not treat a missing field as Ready. Avoid an invented completed/working/idle state machine or percentage.

Implement a rebuildable read projection under the existing SQLite transaction authority. Maintain projection rows and any supporting per-result review aggregates from the same committed mutations that update records/changes. The projection is never accepted as mutation authority. Do not compute it by deserializing every task's complete history per list request, and do not issue one detail query per row. Page query uses an indexed `(session,task_id)` projection scan, capped at 20 rows and 128 KiB encoded. Supporting indexes must make latest-event and exact-submission review lookups bounded. Define review recency by committed event cursor, not UUID ordering or client clock.

For existing databases, rebuild in bounded batches while capability remains absent; verify the rebuilt cursor against the authoritative changes watermark before enabling. Rebuild errors keep the old task-list endpoint usable. An additive materialized projection is acceptable maintenance cost only with a full recompute equivalence test and transaction rollback tests. If an existing projection facility is present by implementation time, extend it rather than creating a parallel table.

Pagination is explicitly snapshot-pinned. Subsequent pages require the first page's session watermark. If relevant session state changed, return revision_conflict; App retains the loaded rows and offers Refresh, never mixes pages. Avoid unrelated sessions invalidating each other's pagination by using a session change watermark. Existing task-list endpoint remains unchanged for compatibility; old Gateways keep neutral title/project rows. Existing changes SSE only raises the updates notice; list refresh uses the summary endpoint. No row-level background GETs and no native polling from the list.

## Accessibility and motion requirements

- Both panes have useful accessibility headings. Rows announce title, project, selected state, and factual attention text in that order. Do not rely on color, checkmarks, or animation. Icons retain explicit labels and controls at least the repository's normal 44/48-point targets.
- Reading order is list then detail on wide layouts; compact inactive panes are inaccessible and unfocusable. Focus moves to the detail heading after an explicit task selection, returns to the original row on Back, and stays stable on new output. Failed network reads announce a concise error without stealing focus from a draft.
- Result acceptance and requested changes controls announce the selected result identity/version context. Recipient controls announce the actual selected attempt. No action label implies a send acknowledgement proves completion.
- Test 200% font scaling and long localized text without fixed-height rows, clipped buttons, or hidden notes. Project paths can visually truncate only when their complete value remains available to assistive technology or explicit detail.
- Reuse `src/lib/motion.ts` system reduced-motion helpers; responsive pane changes and anchor restoration are instantaneous under reduced motion. Avoid pulsing attention indicators and autoanimated reordering. List order remains pinned until explicit refresh.
- Verify TalkBack and hardware keyboard traversal on Android. VoiceOver, iPad multitasking, floating keyboard, and iOS safe-area behavior remain explicitly unverified until the user's iOS testing environment is available.

## Executable ownership and validation sequence

1. Gateway projection owner: new `src/work/store/summaries.rs`, model/HTTP/schema and capability wiring, SQLite migration/indexes, isolated store/service tests. Coordinate shared `work_schema.rs` and capability files before editing. Tests: exact latest submission review, accepted older submission with newer unreviewed result, released ambiguous generation, uncertain reconciliation excluded, foreign session/auth, rollback, rebuild equivalence, bounded row/byte query, pagination conflict, no per-row native/detail calls.
2. App presentation-state owner: new `src/lib/work-view-memory.ts` and focused tests; narrow controller navigation/cache changes; one existing output snapshot owner reused. Test Back/return, rapid task selection, resize without requests or POSTs, stale response rejection, historical selection, pairing rename/re-pair, bounded eviction, pending journal unaffected, and newer draft preservation.
3. App layout owner: new `src/components/work-task-layout.tsx`, narrowly refactor workspace into controlled list/detail/creation regions, add measured dock using existing keyboard animation primitive. Do not alter generic SettingsSheet or terminal delivery behavior. Test width/font breakpoint logic and anchor restore cancellation in pure helpers; native evidence is required for clipping and keyboard geometry.
4. App summary owner: decoder/client list method, compact row view, controller summary snapshot mapping and demo protocol fixture. Preserve fallback endpoint. Tests reject foreign IDs, invalid counts, mismatched revisions/cursors and review references; one list request per page, no N+1 calls; SSE never mutates displayed row order or detail/result/output snapshot.
5. Native QA owner (parent only while device access is serialized): add full-tag task-layout `.ad` flow and extend manifest. Cover phone Back return at historical result, keyboard show/hide and send accessibility, long text, safe-area hit targets, wide layout selection and resize while output arrives, and queued/unconfirmed mutation guards. Reuse the dedicated QA AVD; do not create another device. Restore font/motion/window settings after every flow, including failure. Platform-specific claims require actual platform runs.
6. Localization owner: extract after source labels settle, translate new messages, compile, and run full tsc/lint/format/unit gates. Parent runs the entire Android suite and real paired delivery regression where delivery/layout integration changes warrant it. No readiness claim from targeted/demo tests alone; iOS remains a documented test gate.

Start with presentation-state preservation before moving layout nodes, then dock/split layout, then summary projection. These are independent of enabling native execution or task inputs and must not relax their capability or receipt gates.

ryuhzk:
## Real Android-to-agent validation update

A real Codex 0.154.0 session has now received two explicit messages from the App's ordinary current-terminal composer through the isolated Gateway. Codex executed the requested commands and the returned output was visible in the Android App:

- Initial command: `echo MUQUN_AI_$((6000+41))`; stdout: `MUQUN_AI_6041`.
- Follow-up: `echo MUQUN_FOLLOWUP_$((6100+7))`; stdout: `MUQUN_FOLLOWUP_6107`.

Both messages were submitted once. The output values were computed by the shell, rather than supplied as literal expected output in the prompt. The test used the dedicated `muqun_collaboration_qa` AVD and the explicitly approved isolated working directory. No additional agent approval prompt appeared. App screenshots and an independent native terminal read agree on the returned output. The server list subsequently reported the agent as idle; that status is not used as evidence of task acceptance or completion.

### Exact scope and remaining gates

This proves ordinary App-to-Gateway-to-existing-Codex delivery and follow-up. It does **not** prove the new managed startup, immutable-instance assignment, input attachment, result-reporting, delegation, or lost-acknowledgement workflows. The older isolated Gateway did not expose the immutable instance identity required by the new assignment picker, so that picker correctly excluded its candidates. Those workflows still require the newer isolated stack and their own tests.

The rebuilt Android APK also reports the patched native Cronet transport marker. A fresh dropped-acknowledgement test is still required before claiming one native POST with no replay.

The last full Android run that exercised the App passed 17/18 flows. A subsequent dedicated-device run was blocked by the exploratory session still holding its lease; that run was allowed to terminate, the session was explicitly closed, and the suite was restarted. The remaining attachment-intent check exposed a real accessibility defect: radio controls lacked native checked state and a visible selected indicator. That defect has been fixed; typecheck, lint, formatting, and 3,876 unit tests pass (2 skipped). A full rerun on the dedicated AVD is in progress. The QA App was explicitly unpaired first, and the isolated Gateway's device grants were verified empty. No iOS runtime coverage is claimed.

Implementation remains local until the required gates pass. The independent keyboard branch is separate and will receive its own Android QA and PR.

ryuhzk:
# Exact-instance interruption: App implementation contract

Full Gateway/native contract and execution plan: https://github.com/osuki-dev/muqun-gateway/pull/27#issuecomment-5662128822

Status: App implementation pending. Native implementation is tested; Gateway store and backend adapters are being implemented locally. Capability remains withheld. This continues the existing task/receipt model and does not introduce a second collaboration system.

## 5. App journal, controller and UI

Extend strict operation/receipt decoding in `src/lib/work-api.ts` for `interrupt_attempt` and its evidence. Add `api.interrupt(taskId, attemptId, body, context)`, scope assertions on response, and receipt-kind recovery. Mutation decoder rejects acknowledged interruptions without exact matching evidence; malformed success remains unconfirmed, preserving the pending key.

Extend `src/lib/work-journal.ts` through an explicit versioned migration to preserve the existing single primary pending record and at most one interruption pending record per server/session scope. Reuse SecureStore/native storage and serialized write queue; no plaintext side storage. Old records migrate losslessly. Both lanes carry task/attempt/expected launch/epoch/key/optional operation ID; journal-write failure means no POST. Scope eviction must never discard unresolved entries, and total record/byte limits remain bounded. Unpairing/account changes retain the existing protected cleanup policy.

Add an interruption-specific controller method rather than passing `execution=true` into the ordinary mutation lock unchanged. It requires connected session, capability, complete exact selected-attempt binding, fresh task revision, and no unresolved interruption. It may coexist with a pending delivery for that same attempt; preserve the original delivery intent and text. An unresolved create/start lacking target identity remains ineligible. Fetch metadata if needed without replacing the reader's output snapshot. Keep each asynchronous completion fenced to the original server/session/task/attempt context so it cannot clear another draft or update a replacement assistant.

Persist interrupt intent before POST. On acknowledged receipt clear only its lane; on ambiguity retain it. Relaunch/reconnect performs only GET receipt/operation lookup. An absent receipt is not permission to POST again. Offer “Check interruption status”; no timer, focus event, SSE notification, reconnect, double-tap or app restart resends the action. Add interruption uncertainty to same-attempt managed-delivery guards without turning all unrelated historical operations into global locks.

Place “Interrupt assistant” beside the selected attempt's output/lifecycle controls in `src/components/work-task-workspace.tsx`, using `WorkTaskDetail`'s existing attempt selection. Identify the selected lead/worker and server; retain visible historical attempts. The action means one explicit cancellation request and requires no trust/approval answer. It remains available while observed working/blocked or task paused when identity is valid; unavailable/old backend gets upgrade guidance. Do not send raw Escape from the UI.

Show “Interrupt requested” after a validated write receipt; “Interruption not confirmed” after uncertainty; “Assistant changed” for stale identity. Keep lifecycle observations separate: only native successful-wait reconciliation can say the process exited or replacement reservation was released. Do not say “Stopped,” “Cancelled successfully,” or “Task complete.” Keep output snapshot and scroll position stable; indicate new output until the user refreshes. Interruption must not navigate to the terminal automatically or clear a follow-up draft.

## 7. Acceptance tests

Gateway tests must assert zero port calls on wrong actor/session/task/epoch/revision, unsupported capability, revoked pairing, full quota and duplicate pending admission. Verify an uncertain prompt does not prevent one exact interrupt, and the original prompt state remains unchanged. Two devices race: one interruption admission; same key replay returns identical operation/evidence; changed identity conflicts. Drop the native ACK after a write: record unconfirmed, restart Gateway, perform GET recovery, and count no second native call. Validate malformed/wrong-ID/wrong-key/two-byte receipts as unconfirmed. Terminal evidence quota remains reserved across failure. Interrupt acknowledgement must not release reservations or submit results.

App tests cover journal v1 migration, both pending lanes surviving process restart, journal failure preventing POST, double taps, navigation/pairing changes during completion, exact receipt parsing, offline/unavailable messaging, paused/working eligibility, uncertain-delivery coexistence, no draft loss, and GET-only recovery. Assert operation-kind additions do not break exhaustive labels/demo reducers or older stored records. Register the native e2e flow with the full tag; do not introduce another test format.

Combined real-process protocol proof uses a fresh isolated Herdr/Gateway session and explicit simulated native executable, checking actual byte count plus durable Gateway receipt. Paired encrypted App proof uses the dedicated QA device, a deliberately lost interruption acknowledgement, kill/relaunch, retained original delivery identity, read-only recovery and no second interrupt POST/native byte. Simulated evidence is labeled simulated; any real AI trust or approval answer requires separate explicit permission. All repository-prescribed Gateway/native/App gates and relevant platform device suites must pass before shipping the new capability.

ryuhzk:
## Patched Android transport: real lost-acknowledgement test passed

The dedicated `muqun_collaboration_qa` Android App was paired to an isolated Gateway using encrypted transport. A private byte-preserving proxy forwarded one task-create POST, waited for the Gateway's committed HTTP 200 response, then deliberately dropped that acknowledgement.

Observed and independently checked:

1. The App displayed **Operation not confirmed** and retained the pending request.
2. Proxy evidence showed exactly **one work POST**; the patched native transport did not resend it.
3. SQLite held one task and zero attempts.
4. After an actual App process restart, reopening Tasks recovered the original pending identity from protected storage and read the committed receipt through GET requests.
5. The recovered task appeared with **No assistant has been started for this task**. Across the full experiment there was still exactly one work POST, one task and zero attempts.

The QA App was then unpaired through its UI. Empty Gateway device grants were verified, the App automation session was closed, and the exact owned proxy/fixture processes were stopped. No production Gateway or active user session was used.

Evidence is local build output: `/tmp/muqun-patched-loss-evidence.json`, `/tmp/muqun-patched-loss-unconfirmed.png`, `/tmp/muqun-patched-loss-recovered.png`. Android APK SHA-256: `67d7fc9dca295d9776865a70f7b3153d4717f6096856b2bf57b0f24731acfb8c`. Pinned Gateway SHA-256: `19b6845ada10094fd4f41a86ae91b7d26106abc312311fc36c6490fa5ba4e7c9`.

This closes the previously observed native automatic POST replay defect for this tested Android task-create/restart path. It does not establish iOS behavior, managed startup, delegation, or interruption delivery; those require their own evidence.

Separately, the combined Gateway formatting, strict Clippy and tests passed locally: 55 library tests and 619 binary tests, with 20 ignored. The new interruption service and conditional lead onboarding are implemented locally; optional capabilities remain withheld pending their direct workflow validation.

Per the requested priority, implementation and direct feature tests now take precedence over rerunning unrelated full-suite failures. The App's remaining full-suite failure was traced to agent-device 0.20.10 discarding Android checked/selected fields; the actual selected checkmark rendered correctly. Its flow now checks exact control IDs and visible selection labels, retaining the immutable input-use assertion. A final full run remains required before ready delivery. The keyboard agent now has the dedicated device for its independent feature QA.

ryuhzk:
## Delegation and interruption: isolated process proof passed

A freshly pinned Gateway and the local bound-identity Herdr build passed the complete simulated-assistant process check. This uses actual native PTYs and actual CLI subprocesses with independently inherited per-attempt context, not mocked service returns. The assistants are explicit simulated ELF executables; this is not a real-AI or App/device coverage claim.

Verified across 24 scoped CLI jobs:

- Paired activation of the exact root coordinator; child creation, startup, initial prompt, follow-up and returned output.
- Exact replay and receipt recovery without an additional child, attempt or initial prompt. The observed initial-prompt echo count remained one across replay and follow-up observations.
- Explicit immutable parent input sharing and a worker result submitted through the real local CLI from its worktree. Mutating the source artifact did not change returned submitted bytes.
- The **same waiting child** was refused while its dependency lacked a result version, updated with an exact immutable result through `set-child-dependencies`, recovered through its original receipt, then explicitly started and delivered to. No replacement child was used to bypass the waiting state.
- Pause blocked new child work while explicit follow-up to an existing child remained governed separately. A worker's reporting token did not grant delegation.
- Exact coordinator exit removed its local control and request-key access, while worker result reporting and paired recovery continued.
- Paired interruption returned a validated Escape receipt with `bytes_written: 1` and `input_disposition: written`. The simulated process observed the actual Escape. Replaying the request produced no second Escape marker; the attempt reservation remained reserved.

The first assistant in each child task is that task's `lead`, independently of the root coordinator. This does not enable child delegation. Onboarding was corrected to explain this, explicit result-version selection, dependency updates, and separate startup/delivery receipts.

Gateway SHA-256: `7e029b6d763850fef135cf1d4c4da498e1819ef1e328249775125a32cffd1d71`.
Herdr SHA-256: `c6af5db505a7b218e480d795573d8df96fec69faf980f5daf17953d42aa03c11`.
Local evidence: `/tmp/gwp-tgm712tk/summary.json`, `evidence.jsonl`, and `proof.py`.

Credential scanning passed. The owned Gateway/native processes exited, the private native socket was removed, and no owned workspace process remained. Production services and user sessions were not used.

Remaining work: App delegation/dependency control integration and direct Android testing; real managed-AI App delivery; the independent keyboard's tmux modifier compatibility fix; final integrated gates. Capabilities remain withheld until their App workflow is proven. The later tmux-only source fix is outside this pinned binary's evidence.

ryuhzk:
## Real managed-startup validation and remaining integration

The dedicated Android QA App paired with an isolated encrypted Gateway candidate and created a managed task in the previously authorized test directory. The canonical installed Codex binary started, displayed its directory trust prompt, and reached its interactive screen after the already authorized trust choice. Gateway persisted an acknowledged start operation with an immutable native launch ID and owner epoch. No task prompt was automatically sent while startup was blocked.

This exposed a real integration gap: native `agent.list` does not yet expose the bound launch identity, while Gateway's ordinary agent decoder derives only legacy launch-alias or conversation identities. Consequently the App cannot match the managed attempt to its running recipient and correctly blocks delivery. The fix is to expose the current runtime's immutable bound identity in native discovery and preserve it through Gateway decoding. Pane IDs, assistant names, and idle status must not substitute for that identity. Real managed prompt delivery and returned output remain unproven until this fix is tested through the App.

The run also found two usability issues: definite unsupported native profiles were reported as generic not-ready errors, and returning to Tasks after failed startup could reopen the creation form. The refusal mapping is now fixed and tested locally; the navigation regression is included in the staged explicit-view/layout integration.

Current validation milestones:

- App integration before the remaining layout work: TypeScript, lint, formatting, and 3,907 unit tests passed; two skipped.
- Gateway after summary projection and refusal mapping: 59 library and 626 binary tests passed; 20 ignored. Strict Clippy, formatting, and offline release build passed.
- Responsive layout, bounded reading-state preservation, and factual task summaries are being integrated. Their host checks do not establish Android geometry or accessibility coverage.
- Independent keyboard draft: https://github.com/osuki-dev/muqun-app/pull/82. Its Android keyboard flow and direct SSH-demo typing passed; full-suite failures and Gateway modifier dependency are documented there.

Implementation remains under local validation. This is not a readiness claim, a full-suite pass, or iOS runtime coverage. Unrelated suite failures are recorded while feature completion remains the priority.

ryuhzk:
## Real managed App-to-Codex workflow verified; seamless reporting refinement

On the dedicated Android QA device, the App created a task, started the installed canonical Codex binary in the authorized isolated directory, and delivered the initial instruction to the exact native launch identity. Codex executed the echo command and returned `MUQUN_MANAGED_7347`. An App follow-up reached that same launch and returned `MUQUN_FOLLOWUP_7459`. Both results were submitted through the scoped Gateway reporting interface. The store contains one acknowledged start and exactly two acknowledged prompt-delivery operations for this task.

The App now reads the real nested Gateway output envelope correctly. New output raised a notice without replacing the reading snapshot; explicit Refresh output displayed the follow-up. Task-list refresh showed the submitted result awaiting review. Accepting the older result affected only that exact submission; the newer result remained unreviewed. Requesting changes on the newer result prepared an explicitly addressed draft and did not send it. The App interruption action persisted the exact one-byte Escape receipt and did not claim task completion.

Real testing found and repaired the missing native discovery identity and strict output-envelope decoding. It also identified two presentation corrections now being integrated: keep mutation errors visible beside the composer, and use immutable result identifiers rather than list-position numbers that change after refresh.

The user requires routine reporting to proceed without approval interruptions. Context lookup and result submission were explicitly authorized during this test, but manually answering prompts is not the intended product mechanism. The implementation now adds a per-launch Codex stdio MCP adapter exposing only task-scoped context, result submission, and result-receipt lookup. It reuses existing Gateway authority, binds a pinned executable, configures only these tools for automatic approval, and preserves the ordinary agent sandbox, user configuration, and unrelated approval policy. No generic shell, arbitrary read, review acceptance, merge, or publication tool is exposed. Actual model-driven testing of this new seamless path remains required.

Current implementation is still under final local validation. This comment does not claim the complete Android suite, seamless MCP model-path validation, or iOS runtime coverage has passed.

ryuhzk:
Implementation code is now pushed in `c48f4e0`.

Included: managed task UI and durable delivery, scoped inputs, immutable result review, delegation and dependencies, lifecycle recovery/interruption, protected pending-operation journals, factual summaries, responsive layout, pinned output, and Android mutation transport fixes. Architecture, interaction rationale, security contracts and executable work packages remain in `docs/`.

Validation: TypeScript, lint and formatting passed; App tests passed (3,938 passed, two skipped). Native E2E syntax/manifest validation passed for 27 flows. See `docs/task-workflow-publication-status.md` for the explicit acceptance boundary.

Companion Gateway: https://github.com/osuki-dev/muqun-gateway/pull/27 (includes the exact-base Herdr integration patch). Independent keyboard implementation: https://github.com/osuki-dev/muqun-app/pull/82.


Published at the owner's request as a draft implementation snapshot. **Do not merge.** The earlier documentation-only description predates this code snapshot.

Real paired Android testing verified actual Codex startup, initial echo, follow-up echo, two scoped result submissions, exact-result review, unsent requested-change draft preparation, and a one-byte interruption acknowledgement. Output stayed pinned until explicit refresh. These checks used an isolated session and the dedicated QA AVD.

Still pending: real model-driven proof of the new automatic scoped MCP reporting path, App managed-agent eligibility integration, final full Android acceptance after integration, and remaining layout/accessibility and recovery cases. iOS runtime testing is deferred to the owner's environment. The earlier Android full run passed 17 of 18 flows; its corrected assertion has not yet been validated in the final expanded suite. This snapshot is not ready to merge or release.

ryuhzk:
Architecture correction: official Herdr only

The owner explicitly rejects maintaining a Herdr fork or patch set. This supersedes the previous comment's proposal to ship `integrations/herdr/` as a required dependency. Do not apply or distribute that patch as the product installation path.

The current draft implementation depends on experimental native APIs (`agent.start_bound`, bound prompt/lifecycle/interruption and reporting configuration). Those dependencies must be removed or replaced before acceptance. Merely deleting the patch file would not make the implementation compatible with official Herdr. The paired tests against the modified binary prove that experiment only; they do not establish official-release compatibility.

Required architecture boundary:

- Muqun owns task records, assignment/review history, permissions, reporting and recovery policy in Gateway, with App as the interaction client.
- Herdr remains an unmodified external backend, accessed through its supported public interfaces. No custom native build, patch maintenance or mandatory upstream contribution.
- Reuse the existing collaboration execution adapter. Audit the actual official API guarantees before selecting the managed execution mechanism; do not invent immutable execution identity or exactly-once delivery from a reusable terminal identifier.
- Where official interfaces cannot supply a required guarantee, either implement it for Gateway-owned processes through a supported provider interface, or explicitly restrict that operation. Gateway ownership of metadata alone cannot eliminate backend race conditions.
- Validate startup, echo, follow-up, reporting, uncertain delivery and recovery against an unmodified official binary on the dedicated QA device before claiming completion.

Keep both PRs in draft and unmerged. The App and Gateway task/history/UI work can be retained, but the native execution dependency requires redesign. The existing patch is a rejected experiment, not the accepted maintenance plan.

ryuhzk:
Herdr patch scope and implementation status

The patch changes Herdr’s source code, not its configuration: 32 files, about 3,450 added lines, including tests and API documentation.

It introduces custom native APIs for instance-bound startup, prompt delivery, output lookup, lifecycle evidence and interruption, plus per-launch Codex MCP reporting configuration and managed-agent eligibility metadata. The current draft's managed execution path depends on those changes. Removing the patch file alone would not make that path work with an unmodified official Herdr release.

The owner explicitly does not want Muqun to maintain a Herdr fork or patch set. Introducing this dependency was an architecture mistake, and the patch-based approach is not an accepted implementation direction. Tests against the modified Herdr binary demonstrate the experiment only; they do not establish compatibility with an official, unmodified release. Previous progress reports overstated readiness in that respect.

Task storage, history, result review, parts of the App UI and the independent keyboard work may be retained, but the managed execution integration requires reassessment against official Herdr interfaces before it can be accepted. No replacement architecture has yet been validated.

At the owner's instruction, development is paused after this explanatory comment. No further code changes, commits, pushes or merges will be performed without further instruction. Keep this PR in draft and unmerged.

